PR #6, opened 2024-08-30
RootRegistry: Add support for setting URI

Arachnid:


TateB (approved):

jefflau (changes requested):
Can we write a failing test for setting uri on a name not owned

Arachnid:
@jefflau PTAL

PR #7, opened 2024-10-01
Eth registrar

jefflau:
Eth registrar on the premise that the commit + registrations happen on L2

Todo:

- [ ] Resolver to be updated for _setRecords
- [ ] setRecords in register functionality
- [ ] Add Grace period
- [ ] Fix tests for Exponential Price Curve (not using ens registry)


contracts/src/registrar/ETHRegistrar.sol
```diff
@@ -0,0 +1,184 @@
...
+        (uint64 currentExpiry, ) = registry.nameData(uint256(labelHash));
+        if (currentExpiry < block.timestamp) {
+            revert NameExpired(labelHash);
+        }
+
+        // Calculate new expiry
+        uint64 expires = currentExpiry + duration;
+
+        // Check if the provided value is sufficient
+        IPriceOracle.Price memory price = rentPrice(label, duration);
```
Arachnid, line 142: Does this just swallow any extra?

contracts/test/registrar/TestExponentialPremiumPriceOracle.ts
```diff
@@ -0,0 +1,237 @@
+// import { loadFixture } from '@nomicfoundation/hardhat-toolbox-viem/network-helpers.js'
```
Arachnid, line 1: Should this be commented out?

Arachnid (changes requested):

makoto:
@jefflau do we still need this PR? If not, shall we close?

PR #11, opened 2025-02-05
Owned resolver

makoto:
This is a port of https://github.com/ensdomains/ens-contracts/blob/staging/contracts/resolvers/OwnedResolver.sol but using https://github.com/ensdomains/verifiable-factory to make the resolver verifiable.
It also uses OZ's OwnableUpgradeable instead of Ownable


contracts/src/resolver/OwnedResolver.sol
```diff
@@ -0,0 +1,61 @@
...
+import {PubkeyResolver} from "@ens/contracts/resolvers/profiles/PubkeyResolver.sol";
+import {TextResolver} from "@ens/contracts/resolvers/profiles/TextResolver.sol";
+import {ExtendedResolver} from "@ens/contracts/resolvers/profiles/ExtendedResolver.sol";
+
+/**
+ * @title OwnedResolver
+ * @dev A simple resolver anyone can use; only allows the owner of a node to set its
+ * address.
+ */
+contract OwnedResolver is
```
mdtanrikulu, line 20: There is one missing bit in here for UUPS compliance. The implementation contract must inherit `UUPSUpgradeable`, [see here.](https://github.com/ensdomains/verifiable-factory/blob/3293069b7309987e3118e5c6228e2a7ca6865721/README.md#requirements)

contracts/src/resolver/OwnedResolver.sol
```diff
@@ -0,0 +1,61 @@
...
+    DNSResolver,
+    InterfaceResolver,
+    NameResolver,
+    PubkeyResolver,
+    TextResolver,
+    ExtendedResolver
+    {
+
+    function initialize(address _owner) public initializer {
+        __Ownable_init(_owner); // Initialize Ownable
```
mdtanrikulu, line 34: UUPS initialization must be added here as in [example implementation](https://github.com/ensdomains/verifiable-factory/blob/3293069b7309987e3118e5c6228e2a7ca6865721/README.md#example-implementation).

contracts/src/resolver/OwnedResolver.sol
```diff
@@ -0,0 +1,61 @@
...
+    }
+
```
mdtanrikulu, line 36: we will also need to add following override to limit implementation upgrades to the owner only.

```solidity
function _authorizeUpgrade(address) internal override onlyOwner {}
```

mdtanrikulu (changes requested):

contracts/src/resolver/OwnedResolver.sol
```diff
@@ -0,0 +1,67 @@
...
+    ExtendedResolver
+    {
+
+    function initialize(address _owner) public initializer {
+        __Ownable_init(_owner); // Initialize Ownable
+        __UUPSUpgradeable_init();
+    }
+
+    function isAuthorised(bytes32) internal view override returns (bool) {
+        return msg.sender == owner();
```
Arachnid, line 42: Can this use `onlyOwner`?

Arachnid (changes requested):

contracts/src/resolver/OwnedResolver.sol
Arachnid: No, I mean use `onlyOwner` as a modifier instead of calling `isAuthorized` inside the relevant functions.

Arachnid (approved):

mdtanrikulu (approved):

PR #12, opened 2025-02-07
refactor: improvement to expiry logic

hiddentao:
- Refactored expiry extractor into re-usable method
- Fixed expiry logic in ownerOf(), getResolver(), etc.
- Burn expired token before minting new one


contracts/src/registry/BaseRegistry.sol
```diff
@@ -35,7 +35,7 @@ abstract contract BaseRegistry is IRegistry, ERC1155Singleton {
     modifier withSubregistryFlags(uint256 tokenId, uint96 mask, uint96 expected) {
         (, uint96 flags) = datastore.getSubregistry(tokenId);
         if (flags & mask != expected) {
-            revert InvalidSubregistryFlags(tokenId, flags & mask, expected);
+            revert InvalidSubregistryFlags(tokenId, flags & mask, expected);
```
Arachnid, line 38:
```suggestion
            revert InvalidSubregistryFlags(tokenId, flags & mask, expected);
```

Arachnid (approved):

PR #13, opened 2025-02-10
refactor: better naming conventions for "locking" mechanism

hiddentao:
* `LockableRegistry` - `UpdatableRegistry`
* `lock()` -> `setFlags()`

### DIFF
--- a/contracts/src/registry/UpdatableRegistry.sol
+++ b/contracts/src/registry/UpdatableRegistry.sol
@@ -9,7 +9,7 @@ import {IRegistry} from "./IRegistry.sol";
 import {IRegistryDatastore} from "./IRegistryDatastore.sol";
 import {BaseRegistry} from "./BaseRegistry.sol";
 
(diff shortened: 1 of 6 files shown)


contracts/src/registry/UpdatableRegistry.sol
Arachnid, line 1: How about `FlagsRegistry` or `PermissionedRegistry`?

Arachnid (changes requested):

contracts/src/registry/UpdatableRegistry.sol
hiddentao: Was thinking Updatable just because it also allows for the subregistry and resolver to be updated.

hiddentao: Took another look at it, and PermissionedRegistry does feel a bit nicer

Arachnid (approved):

PR #15, opened 2025-02-11
feat: first attempt at v2 ETH registrar

hiddentao:


contracts/src/registry/ETHRegistrar.sol
```diff
@@ -0,0 +1,52 @@
...
+
+import {IETHRegistrar} from "./IETHRegistrar.sol";
+import {IRegistry} from "./IRegistry.sol";
+import {IERC1155Singleton} from "./IERC1155Singleton.sol";
+import {IETHRegistry} from "./IETHRegistry.sol";
+
+contract ETHRegistrar is IETHRegistrar {
+    IETHRegistry public registry;
+    mapping(address => bool) public controllers;
+    mapping(uint256 => bool) public registered;
```
Arachnid, line 12: Is this necessary?

contracts/src/registry/ETHRegistrar.sol
```diff
@@ -0,0 +1,52 @@
...
+pragma solidity >=0.8.13;
+
+import {IETHRegistrar} from "./IETHRegistrar.sol";
+import {IRegistry} from "./IRegistry.sol";
+import {IERC1155Singleton} from "./IERC1155Singleton.sol";
+import {IETHRegistry} from "./IETHRegistry.sol";
+
+contract ETHRegistrar is IETHRegistrar {
+    IETHRegistry public registry;
+    mapping(address => bool) public controllers;
```
Arachnid, line 11: We can probably use OZ's roles pattern here to enable controllers instead.

contracts/src/registry/IRegistryController.sol
```diff
@@ -0,0 +1,51 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.8.13;
+
+import {AccessControl} from "@openzeppelin/contracts/access/AccessControl.sol";
+
+interface IRegistryController {
```
Arachnid, line 6: Most of this functionality looks like it's something the registrar would have rather than a controller? I'm not sure we need an interface for the controller at all, since it only calls the registrar, it doesn't get called by it.

Arachnid (changes requested):

contracts/src/registry/ETHRegistrar.sol
hiddentao: Apologies, was planning to remove this but then I got caught up in a merge.

contracts/src/registry/IRegistryController.sol
hiddentao: Apologies, this was supposed to be deleted.

contracts/src/registry/ETHRegistrar.sol
hiddentao: Nice idea, will do

makoto:
Is this now obsolete? https://github.com/ensdomains/namechain/pull/7/files

makoto:
Shouldn't EthRegistrar now be called EthRegistryController? 
According to the design doc https://drive.google.com/file/d/1LoA-xboajzUuWwsWpxJ-h1-WqDEsRlkC/view

```
4.5.3 L2 .eth Registry Controller

ENS v2 Design Doc 16
Similar to the registrar controller in ENS v1, the .eth registry controller will exist
on L2 and facilitate registration and renewals of names by users. As in v1, this
division exists in order to ensure registration and renewal logic, including
pricing, can be updated, while allowing the .eth registry itself to be locked.
Multiple controllers can be authorised by the DAO as needed.
```

contracts/src/registry/ETHRegistrar.sol
Arachnid, line 1: I'm a bit confused, because now this just seems to pass through to the registry without doing anything of its own. If this is intended to be the controller that users call, then we should remove the access restriction and add things like taking payment and having a price oracle.

Arachnid (changes requested):

contracts/src/registry/ETHRegistrar.sol
hiddentao: This is intended as an initial base for those features you've mentioned. I was assuming we'd do the price oracle work separately and then integrate that into this. I can add a TODO comment to this as a sort of placeholder?

makoto:
Don't we need to implement `commit` & `register` on the registry controller?

contracts/src/registry/ETHRegistrar.sol
Arachnid: In v1, we have the registrar and the registrar controller, but since v2 permits arbitrary registry controller implementations, the registrar role can be combined into the registry. The v1 registrar controller can then either be renamed to 'registrar' (since it's what users interface with to register) or left as-is; I personally prefer the former.

Right now I think we already have the core of the registrar functionality in the new .eth registry, so this PR could make a start on the controller instead.

Arachnid:
Yes, the controller will still need to have a commit-reveal implementation.

hiddentao:
@Arachnid Got the commit-reveal in there. As for the v1 reverse resolver stuff, should I put that in too?

contracts/src/registry/ETHRegistrar.sol
```diff
@@ -0,0 +1,214 @@
...
+    error DurationTooShort(uint64 duration);
+    error CommitmentTooNew(bytes32 commitment);
+    error CommitmentTooOld(bytes32 commitment);
+    error NameNotAvailable(string name);
+    error InsufficientValue();
+
+    IETHRegistry public immutable registry;
+    IPriceOracle public immutable prices;
+    uint256 public immutable minCommitmentAge;
+    uint256 public immutable maxCommitmentAge;
```
Arachnid, line 29: We should consider making these mutable, since the contract already has an owner.

contracts/src/registry/ETHRegistrar.sol
```diff
@@ -0,0 +1,214 @@
...
+    }
+
+    /**
+     * @dev Check if a name is available.
+     * @param name The name to check.
+     * @return True if the name is available, false otherwise.
+     */
+    function available(string calldata name) external view returns (bool) {
+        address subregistry = address(registry.getSubregistry(name));
+        return subregistry == address(0);
```
Arachnid, line 62: We want to permit users to set their subregistry to 0 for valid names if they don't have subdomains. We should probably check the expiration date, instead.

contracts/src/registry/ETHRegistrar.sol
```diff
@@ -0,0 +1,214 @@
...
+
+    /**
+     * @dev Commit a commitment.
+     * @param commitment The commitment to commit.
+     */
+    function commit(bytes32 commitment) public override {
+        if (commitments[commitment] + maxCommitmentAge >= block.timestamp) {
+            revert UnexpiredCommitmentExists(commitment);
+        }
+        commitments[commitment] = block.timestamp;
```
Arachnid, line 118: We should emit an event when a commitment is recorded.

contracts/src/registry/ETHRegistrar.sol
```diff
@@ -0,0 +1,214 @@
...
+     * @param owner The owner of the name.
+     * @param subregistry The subregistry to register the name in.
+     * @param flags The flags to set on the name.   
+     * @param duration The duration of the registration.
+     * @return tokenId The token ID of the registered name.
+     */
+    function register(
+        string calldata name,
+        address owner,
+        IRegistry subregistry,
```
Arachnid, line 135: It'd be useful to allow setting the resolver at the same time as registering.

contracts/src/registry/ETHRegistrar.sol
```diff
@@ -0,0 +1,214 @@
...
+        uint96 flags,
+        uint64 duration
+    ) external payable onlyRole(CONTROLLER_ROLE) returns (uint256 tokenId) {
```
Arachnid, line 138:
```suggestion
    ) external payable returns (uint256 tokenId) {
```
I think we want this to be publicly callable now, right?

contracts/src/registry/ETHRegistrar.sol
```diff
@@ -0,0 +1,214 @@
...
+
+        emit NameRegistered(name, owner, subregistry, flags, duration, tokenId);
+    }
+
+    /**
+     * @dev Renew a name.
+     * @param name The name to renew.
+     * @param duration The duration of the renewal.
+     */
+    function renew(string calldata name, uint64 duration) external payable onlyRole(CONTROLLER_ROLE) {
```
Arachnid, line 161:
```suggestion
    function renew(string calldata name, uint64 duration) external payable {
```

contracts/src/registry/ETHRegistrar.sol
```diff
@@ -0,0 +1,214 @@
...
+
+        (uint64 expiry, ) = registry.nameData(tokenId);
+
+        registry.renew(tokenId, expiry + duration);
+
+        if (msg.value > totalPrice) {
+            payable(msg.sender).transfer(msg.value - totalPrice);
+        }
+
+        emit NameRenewed(name, duration, tokenId);
```
Arachnid, line 178: This should include the new expiration date.

Arachnid (changes requested):

Arachnid:
> As for the v1 reverse resolver stuff, should I put that in too?
No, we'll use batch calls and smart accounts to handle that.

contracts/src/registry/ETHRegistrar.sol
hiddentao: If it's publicly callable what's the purpose of the controler role then?

contracts/src/registry/ETHRegistrar.sol
hiddentao: If it's publicly callable what's the purpose of the controler role then?

contracts/src/registry/ETHRegistrar.sol
hiddentao: I don't understand this comment. Unless we provide methods to change these values, what's the point of making them mutable?

contracts/src/registry/ETHRegistrar.sol
Arachnid: There isn't one - the registry needs a controller, which will be this contract. This contract doesn't need a controller role because it's intended to be called by end users.

Arachnid (changes requested):

PR #16, opened 2025-02-16
feat: registry metadata interface

hiddentao:


hiddentao (commented):
Wasn't sure whether metadata generator needed to be replaceable or not, so have gone for it not being so. Also, for the RootRegistry I just override the uri() method as before for efficiency sake. Thus, metadata interfaces only really apply for UserRegistry.

contracts/src/registry/BaseRegistry.sol
```diff
@@ -12,16 +12,18 @@ import {ERC165} from "@openzeppelin/contracts/utils/introspection/ERC165.sol";
 import {ERC1155Singleton} from "./ERC1155Singleton.sol";
 import {IRegistryDatastore} from "./IRegistryDatastore.sol";
 import {IRegistry} from "./IRegistry.sol";
-
+import {IRegistryMetadata} from "./IRegistryMetadata.sol";
```
Arachnid, line 15:
```suggestion
import {IRegistryMetadata} from "./IRegistryMetadata.sol";

```

contracts/src/registry/IRegistryMetadata.sol
```diff
@@ -0,0 +1,14 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.8.13;
+
+/**
+ * @dev Interface for the ENSv2 registry metadata, which stores the token URI for each registry.
```
Arachnid, line 5:
```suggestion
 * @dev Interface for providing metadata URIs for ENSv2 registry contracts.
```

contracts/src/registry/IRegistryMetadata.sol
```diff
@@ -0,0 +1,14 @@
...
+ */
+interface IRegistryMetadata {
+    /**
+     * @dev Fetches the token URI for a node.
+     * @param tokenId The ID of the node to fetch a URI for.
+     * @return The token URI for the node.
```
Arachnid, line 11:
```suggestion
     * @dev Fetches the token URI for a token ID.
     * @param tokenId The ID of the token to fetch a URI for.
     * @return The token URI for the token.
```

contracts/src/registry/BaseRegistry.sol
Arachnid, line 1: I don't think this should be a feature of the BaseRegistry; some registry implementations may want to implement their own metadata URI support.

Arachnid (changes requested):

Arachnid:
It would also make sense to have an implementation that returns the same URL for all token IDs (useful since 1155 permits token ID substitution strings).

contracts/src/registry/BaseRegistry.sol
hiddentao: They already do by overriding these methods.

hiddentao:
@Arachnid I've added another implementation for supporting base URIs, like in erc1155.

hiddentao:
@Arachnid Ok, I've refactored metadata provision out into a mixin. I think this works quite well for now.

contracts/src/registry/MetadataMixin.sol
```diff
@@ -0,0 +1,28 @@
...
+pragma solidity ^0.8.17;
+
+import "./IRegistryMetadata.sol";
+
+/// @title MetadataMixin
+/// @notice Mixin contract for Registry implementations to delegate metadata to an external provider
+/// @dev Inherit this contract to add metadata functionality to Registry contracts
+abstract contract MetadataMixin {
+    /// @notice The metadata provider contract
+    IRegistryMetadata public immutable metadataProvider;
```
Arachnid, line 11: I think this probably needs to be mutable, with a role and a method to change it.

Arachnid (changes requested):
One last change!

contracts/src/registry/MetadataMixin.sol
hiddentao: I want to wait until after the new EAC stuff is in place before assigning a role. Plus, perhaps some registries will want to re-use an existing role for this. So for now I'm just going to add an internal method to update the value.

Arachnid (approved):

PR #17, opened 2025-02-18
feat: migration of fuses to v2

hiddentao:
Notes:  

* I started out by putting incorporating the new flags into the existing flags architecture but things quickly got complicated so for now I’m just keeping it inside NameWrapperRegistry - will revisit later 
* The PARENT_CANNOT_CONTROL fuse is matched by locking the parent registry, meaning: If we migrate a v1 name wrapper where a few names are locked then in v2 the migrated version would be such that all the names are locked (i.e. the subregistry itself is locked and can’t be changed). I think this is reasonable. 
* CANNOT_CREATE_SUBDOMAIN can be handled by removing the registrar role 
* CANNOT_SET_RESOLER can be handled using the existing flags system

### DIFF
--- a/contracts/src/registry/NameWrapperRegistry.sol
+++ b/contracts/src/registry/NameWrapperRegistry.sol
@@ -1,0 +1,217 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.8.13;
+
+import {AccessControl} from "@openzeppelin/contracts/access/AccessControl.sol";
+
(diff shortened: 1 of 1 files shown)


Arachnid (changes requested):
Can you provide a more detailed design rationale? I'm not really understanding why we need to have registrar-controlled settings (and is this intended to mean settings controlled by a third party designated as the registrar, or settings controlled by the owner of the parent name?

PR #18, opened 2025-02-23
fix: ensure there is only one flags storage value for a given label

hiddentao:
Registry and and resolver values for a label had different flags set, even though at the higher level there is only one set of flags. This aligns the registry datastore to be the same.


Arachnid:
I'm not sure this is a good idea. It adds extra overhead to storage and retrieval, without adding new functionality. Presently we expose flags on each because those bits are unused; it's up to implementations to decide if they want to use them for something or not.

hiddentao:
Ok, will close.

PR #19, opened 2025-02-23
feat: use context-based role assignments to implement the permission model

hiddentao:
Notes:

- All new ACL that allows for context-based role setting and other fun features. Fully tested!
- The transfer lock is complex to do using roles because it would require going into the ERC1155 too to handle approvals etc. So I opted for a flag instead for that one.


contracts/src/registry/EnhancedAccessControl.sol
```diff
@@ -0,0 +1,220 @@
...
+
+    /**
+     * @dev Returns `true` if `account` has been granted `role`.
+     */
+    function hasRole(bytes32 context, bytes32 role, address account) public view virtual returns (bool) {
+        uint256 roleVersion = _roleVersion[context][role];
+        uint256 roleVersionInRoot = _roleVersion[ROOT_CONTEXT][role];
+
+        return (_roles[context][role][account].hasRole && _roles[context][role][account].version == roleVersion) ||
+               (_roles[ROOT_CONTEXT][role][account].hasRole && _roles[ROOT_CONTEXT][role][account].version == roleVersionInRoot);
```
Arachnid, line 82: This could be made more efficient by only checking the `roleVersionInRoot` mapping if the check for a direct role fails.

contracts/src/registry/EnhancedAccessControl.sol
```diff
@@ -0,0 +1,220 @@
...
+     * @dev See {IERC165-supportsInterface}.
+     */
+    function supportsInterface(bytes4 interfaceId) public view virtual override returns (bool) {
+        return interfaceId == type(EnhancedAccessControl).interfaceId || super.supportsInterface(interfaceId);
+    }
+
+    /**
+     * @dev Returns `true` if `account` has been granted `role`.
+     */
+    function hasRole(bytes32 context, bytes32 role, address account) public view virtual returns (bool) {
```
Arachnid, line 77: Perhaps an overload with `context` set to `ROOT_CONTEXT` can be provided for convenience?

contracts/src/registry/EnhancedAccessControl.sol
```diff
@@ -0,0 +1,220 @@
...
+     * @dev Attempts to grant `role` to `account` and returns a boolean indicating if `role` was granted.
+     *
+     * Internal function without access restriction.
+     *
+     * May emit a {RoleGranted} event.
+     */
+    function _grantRole(bytes32 context, bytes32 role, address account) internal virtual returns (bool) {
+        if (!hasRole(context, role, account)) {
+            _roles[context][role][account].hasRole = true;
+            _roles[context][role][account].version = _roleVersion[context][role];
```
Arachnid, line 196: Worth verifying, but I believe that both variables will be packed into a single slot, and a setter using a struct literal will thus be more efficient.

contracts/src/registry/EnhancedAccessControl.sol
```diff
@@ -0,0 +1,220 @@
...
+    }
+
+    /**
+     * @dev Returns the admin role that controls `role`. See {grantRole} and
+     * {revokeRole}.
+     *
+     * To change a role's admin, use {_setRoleAdmin}.
+     */
+    function getRoleAdmin(bytes32 role) public view virtual returns (bytes32) {
+        return _adminRoles[role];
```
Arachnid, line 110: Shouldn't this have a fallthrough case for the default admin role?

contracts/src/registry/EnhancedAccessControl.sol
Arachnid, line 1: One issue I see with this approach is that you need to separately configure each role that an account has. Perhaps a solution using a bitmap of roles available for each `(account, context)` would be more flexible and efficient here.

contracts/src/registry/EnhancedAccessControl.sol
```diff
@@ -0,0 +1,220 @@
...
+
+pragma solidity ^0.8.20;
+
+import {Context} from "@openzeppelin/contracts/utils/Context.sol";
+import {ERC165} from "@openzeppelin/contracts/utils/introspection/ERC165.sol";
+
+/**
+ * @dev An enhanced version of OpenZeppelin's AccessControl that allows for:
+ * 
+ * - Context-based roles.
```
Arachnid, line 12: Rather than calling this 'context' perhaps call it 'principal'?

contracts/src/registry/NameWrapperRegistry.sol
```diff
@@ -0,0 +1,247 @@
...
+
+    function ownerOf(uint256 tokenId)
+        public
+        view
+        virtual
+        override(ERC1155Singleton, IERC1155Singleton)
+        returns (address)
+    {
+        (, uint96 oldFlags) = datastore.getSubregistry(tokenId);
+        uint64 expires = _extractExpiry(oldFlags);
```
Arachnid, line 48: You can drop the 'old'.

contracts/src/registry/NameWrapperRegistry.sol
```diff
@@ -0,0 +1,247 @@
...
+            _burn(previousOwner, tokenId, 1);
+        }
+
+        _mint(owner, tokenId, 1, "");
+        datastore.setSubregistry(tokenId, address(registry), flags);
+
+        // registrar has some permissions by default
+        bytes32 tokenIdContext = _tokenIdContext(tokenId);
+        _grantRole(tokenIdContext, RENEW_ROLE, msg.sender);
+        _grantRole(tokenIdContext, EMANCIPATOR_ROLE, msg.sender);
```
Arachnid, line 81: This seems dangerous insofar as changing the registrar won't change who has control over existing names, which may be an unintended gotcha. What about using the admin fallthrough and granting these roles in the root context to the registrar account(s)? This would also make it easier to revoke this right from all domains at once.

contracts/src/registry/NameWrapperRegistry.sol
```diff
@@ -0,0 +1,247 @@
...
+    /** 
+     * @dev Emancipate the owner of the token. 
+     * 
+     * See v1 fuse: PARENT_CANNOT_CONTROL
+     */
+    function emancipate(uint256 tokenId) public onlyRole(_tokenIdContext(tokenId), EMANCIPATOR_ROLE) {
+        bytes32 tokenIdContext = _tokenIdContext(tokenId);
+        _grantRole(tokenIdContext, EMANCIPATED_OWNER_ROLE, ownerOf(tokenId));
+        renounceRole(tokenIdContext, EMANCIPATOR_ROLE, msg.sender); // remove the emancipator role since we can only do this once.
+    }
```
Arachnid, line 112: I freely admit I haven't thought all the implications of this through, but do we need to have this as an explicit thing? Can't we just revoke our own control over the name while giving the same control to the owner of the subname?

contracts/src/registry/NameWrapperRegistry.sol
```diff
@@ -0,0 +1,247 @@
...
+    
+    /**
+     * @dev Allow the owner to renew the name.
+     * 
+     * See v1 fuse: CAN_EXTEND_EXPIRY
+     */
+    function allowOwnerToRenew(uint256 tokenId) public onlyRole(_tokenIdContext(tokenId), EMANCIPATOR_ROLE) {
```
Arachnid, line 119: Does this mean that if you emancipate a name, it's no longer possible to grant the owner permission to renew it?

contracts/src/registry/NameWrapperRegistry.sol
```diff
@@ -0,0 +1,247 @@
...
+        bytes32 tokenIdContext = _tokenIdContext(tokenId);
+        _grantRole(tokenIdContext, RENEW_ROLE, ownerOf(tokenId));
+    }
+
+
+    /**
+     * @dev Prevent the name from being renewed.
+     * 
+     * See v1 fuse: CANNOT_SET_TTL
```
Arachnid, line 128: TTL is actually a separate concept that was never used in ENS. This can be removed entirely.

contracts/src/registry/NameWrapperRegistry.sol
```diff
@@ -0,0 +1,247 @@
...
+        bytes32 tokenIdContext = _tokenIdContext(tokenId);
+        _revokeRoleAssignments(tokenIdContext, RENEW_ROLE);
+    }
+
+    /**
+     * @dev Prevent the name from being transferred.
+     * 
+     * See v1 fuse: CANNOT_TRANSFER
+     */
+    function lockTransfers(uint256 tokenId) public onlyRole(_tokenIdContext(tokenId), EMANCIPATED_OWNER_ROLE) {
```
Arachnid, line 140: Are these functions presently only available on emancipated names?

Arachnid: I'm not actually sure we need this. The sole right that owning the NFT gives you is the right to transfer it, so if you want to block transfers, you could simply transfer the name to a burn address. All other functionality would continue to work fine!

contracts/src/registry/NameWrapperRegistry.sol
```diff
@@ -0,0 +1,247 @@
...
+        bytes32 tokenIdContext = _tokenIdContext(tokenId);
+        transferLock[tokenIdContext] = true;
+    }
+
+    /**
+     * @dev Renounce the emancipated owner role.
+     * 
+     * See v1 fuse: CANNOT_BURN_FUSES
+     */
+    function renounceEmancipatedOwnerRole(uint256 tokenId) public onlyRole(_tokenIdContext(tokenId), EMANCIPATED_OWNER_ROLE) {
```
Arachnid, line 150: There's several specialized functions like this; can we make a generic 'renounceRole' function instead?

contracts/src/registry/NameWrapperRegistry.sol
```diff
@@ -0,0 +1,247 @@
...
+    function supportsInterface(bytes4 interfaceId) public view override(BaseRegistry, EnhancedAccessControl) returns (bool) {
+        return interfaceId == type(IRegistry).interfaceId || super.supportsInterface(interfaceId);
+    }
+
+    function getSubregistry(string calldata label) external view virtual override returns (IRegistry) {
+        (address subregistry, uint96 flags) = datastore.getSubregistry(uint256(keccak256(bytes(label))));
+        uint64 expires = _extractExpiry(flags);
+        if (expires <= block.timestamp) {
+            return IRegistry(address(0));
+        }
```
Arachnid, line 188: I see this reused in several places; we should use an accessor function to remove the duplication.

Arachnid (changes requested):

contracts/src/registry/NameWrapperRegistry.sol
hiddentao: Good point!

contracts/src/registry/NameWrapperRegistry.sol
hiddentao: I did consider role-transference but this would violate the rule that a role can only be assigned someone who is allowed to assign it (e.g admin can assign registrar role). If a user can transfer their role to someone else then they are effectively able to bypass that.

contracts/src/registry/NameWrapperRegistry.sol
hiddentao: This is a bug, will fix.

contracts/src/registry/NameWrapperRegistry.sol
hiddentao: I was just replicate fuses, but sure, will remove it.

contracts/src/registry/NameWrapperRegistry.sol
hiddentao: You're right, there's already a method for that.

contracts/src/registry/NameWrapperRegistry.sol
hiddentao: Because we want to be able to emancipate one name at a time, no?

contracts/src/registry/EnhancedAccessControl.sol
hiddentao: Why? I prefer context as that makes it clearer.

contracts/src/registry/EnhancedAccessControl.sol
hiddentao: Default admin role is 0x00 so fallthrough is automatic.

contracts/src/registry/EnhancedAccessControl.sol
hiddentao: Not sure I understand. We have 3 dimensions (context, role, user) with each dimension's range being effectively infinite. If we use a bitmap we'd have to restrict one or more of the dimensions in some way. Also, the way the structures are setup now is to allow for all actions to be O(1).

Arachnid: Right, I'm suggesting limiting the roles to at most 256 by using a bitmap. My point is that at the moment if you want to add a new user with broad permissions, you have to make one operation per role you want to grant. If you want to transfer roles to a new account, that doubles. With a bitmask of roles, it's only 1 operation in the former case and 2 in the latter.

contracts/src/registry/EnhancedAccessControl.sol
Arachnid: "Context" is very generic and collides with the use of it in OZ above. How about "resource"?

contracts/src/registry/NameWrapperRegistry.sol
Arachnid: I think we could still do this, since rights applied to individual resources override those applied globally.

contracts/src/registry/NameWrapperRegistry.sol
Arachnid: The parent owner cannot - but if I'm reading this correctly, the emancipated owner cannot renew either!

contracts/src/registry/EnhancedAccessControl.sol
Arachnid: I think this would also eliminate the need for role groups.

hiddentao:
@Arachnid Given that we only need to transfer two roles to a new registrar (registrar and renew) I felt that it was easier to do that manually whilst keeping the current storage approach in `EnhancedAccessControl`, i.e. not using a bitmap. Even if we use a Bitmap then keeping the "revoke all assignments of this role within this resource" O(1) feature would still require versioning of assignments, meaning that we'd still be doing a lot of complex storage work in access control. 

Plus, role transfers only work across a particular context. There's no way to transfer all of a user's roles in all contexts. I suppose if we implement hierarchical contexts of some sort we could perhaps do this. But again, this complicates the access control storage.

Happy to discuss this further over calls.

contracts/src/registry/BaseRegistry.sol
```diff
@@ -74,7 +74,7 @@ abstract contract BaseRegistry is IRegistry, ERC1155Singleton {
      * @return The address of the registry for this subdomain, or `address(0)` if none exists.
      */
     function getSubregistry(string calldata label) external view virtual returns (IRegistry) {
-        (address subregistry,) = datastore.getSubregistry(uint256(keccak256(bytes(label))));
+        (address subregistry,) = datastore.getSubregistry(_labelToTokenId(label));
```
Arachnid, line 77: I saw a library function for this in one of your other PRs - can we use that?

contracts/src/registry/EnhancedAccessControl.sol
```diff
@@ -0,0 +1,302 @@
+// SPDX-License-Identifier: MIT
+// OpenZeppelin Contracts (last updated v5.0.0) (access/AccessControl.sol)
+
+pragma solidity ^0.8.20;
+
+import {Context} from "@openzeppelin/contracts/utils/Context.sol";
+import {ERC165} from "@openzeppelin/contracts/utils/introspection/ERC165.sol";
+
+/**
+ * @dev An enhanced version of OpenZeppelin's AccessControl that allows for:
```
Arachnid, line 10: I'm not sure this qualifies as a version of OZ's AccessControl at this point!

contracts/src/registry/EnhancedAccessControl.sol
```diff
@@ -0,0 +1,302 @@
...
+    error EnhancedAccessControlBadConfirmation();
+    error EnhancedAccessControlLockedRole(bytes32 resource, uint256 role);
+
+    event EnhancedAccessControlRoleAdminChanged(uint256 role, uint256 previousAdminRole, uint256 newAdminRole);
+    event EnhancedAccessControlRoleGranted(bytes32 resource, uint256 role, address account, address sender);
+    event EnhancedAccessControlRolesGranted(bytes32 resource, uint256 roleBitmap, address account, address sender);
+    event EnhancedAccessControlRolesCopied(bytes32 srcResource, bytes32 dstResource, address account, address sender, uint256 roleBitmap);
+    event EnhancedAccessControlRolesRevoked(bytes32 resource, uint256 role, address account, address sender);
+    event EnhancedAccessControlAllRolesRevoked(bytes32 resource, address account, address sender);
+    event EnhancedAccessControlRolesLocked(bytes32 resource, uint256 roleBitmap);
```
Arachnid, line 28: Maybe we can abbreviate the prefix to `EAC`.

contracts/src/registry/EnhancedAccessControl.sol
```diff
@@ -0,0 +1,302 @@
...
+
+    /** 
+     * @dev admin role that controls a given role. 
+     * RoleId -> AdminRoleId
+     */
+    mapping(uint256 role => uint256 adminRole) private _adminRoles;
```
Arachnid, line 34: Can we simply say that the upper 128 bits are admin roles and the lower 128 bits are user roles, and thus remove the need to map them explicitly?

contracts/src/registry/EnhancedAccessControl.sol
```diff
@@ -0,0 +1,302 @@
...
+     * @dev user roles within a resource stored as a bitmap.
+     * Resource -> User -> RoleBitmap
+     */
+    mapping(bytes32 resource => mapping(address account => uint256 roleBitmap)) private _roles;
+
+    /**
+     * @dev locked roles within a resource stored as a bitmap.
+     * Resource -> LockedRoleBitmap
+     */
+    mapping(bytes32 resource => uint256 lockedRoleBitmap) private _lockedRoles;
```
Arachnid, line 46: Is this necessary? If a role for a given resource has no admins, it's locked implicitly.

contracts/src/registry/EnhancedAccessControl.sol
```diff
@@ -0,0 +1,302 @@
...
+     */
+    function supportsInterface(bytes4 interfaceId) public view virtual override returns (bool) {
+        return interfaceId == type(EnhancedAccessControl).interfaceId || super.supportsInterface(interfaceId);
+    }
+
+    /**
+     * @dev Returns `true` if `account` has been granted all roles in the root resource.
+     */
+    function hasRootRoles(uint256 rolesBitmap, address account) public view virtual returns (bool) {
+        return (_roles[ROOT_RESOURCE][account] & rolesBitmap) != 0;
```
Arachnid, line 105:
```suggestion
        return (_roles[ROOT_RESOURCE][account] & rolesBitmap) == rolesBitmap;
```

As it is it checks if the user has *any* of the provided roles, but I think we want to check if the user has *all* of the provided roles.

contracts/src/registry/EnhancedAccessControl.sol
```diff
@@ -0,0 +1,302 @@
...
+    }
+
+    /**
+     * @dev Returns `true` if `account` has been granted all roles in `resource`.
+     */
+    function hasRoles(bytes32 resource, uint256 rolesBitmap, address account) public view virtual returns (bool) {
+        return ((_roles[resource][account] & rolesBitmap) != 0) || hasRootRoles(rolesBitmap, account);
```
Arachnid, line 112:
```suggestion
        return ((_roles[resource][account] | _roles[ROOT_RESOURCE][account]) & rolesBitmap) == rolesBitmap;
```

As above - but we need to OR together their resource and root roles to check if they have all required permissions across both scopes.

contracts/src/registry/EnhancedAccessControl.sol
```diff
@@ -0,0 +1,302 @@
...
+     * If `account` had not been already granted `role`, emits a {RoleGranted}
+     * event.
+     *
+     * Requirements:
+     *
+     * - the caller must have ``role``'s admin role.
+     *
+     * May emit a {RoleGranted} event.
+     */
+    function grantRole(bytes32 resource, uint256 role, address account) public virtual canGrantRole(resource, role) returns (bool) {
```
Arachnid, line 145:
```suggestion
    function grantRoles(bytes32 resource, uint256 role, address account) public virtual canGrantRole(resource, role) returns (bool) {
```

contracts/src/registry/EnhancedAccessControl.sol
```diff
@@ -0,0 +1,302 @@
...
+     *
+     * If `account` had been granted `role`, emits a {RoleRevoked} event.
+     *
+     * Requirements:
+     *
+     * - the caller must have ``role``'s admin role.
+     *
+     * May emit a {RoleRevoked} event.
+     */
+    function revokeRole(bytes32 resource, uint256 role, address account) public virtual canGrantRole(resource, role) returns (bool) {
```
Arachnid, line 161:
```suggestion
    function revokeRoles(bytes32 resource, uint256 role, address account) public virtual canGrantRole(resource, role) returns (bool) {
```

contracts/src/registry/EnhancedAccessControl.sol
```diff
@@ -0,0 +1,302 @@
...
+        *
+     * Returns `true` if the role was revoked, `false` otherwise.
+     */
+    function renounceRole(bytes32 resource, uint256 role, address callerConfirmation) public virtual returns (bool) {
+        if (callerConfirmation != _msgSender()) {
+            revert EnhancedAccessControlBadConfirmation();
+        }
+
+        return _revokeRoles(resource, role, callerConfirmation);
+    }
```
Arachnid, line 190: I think we should omit this; if an account is an admin for the role, they can revoke both the role and their own admin status, but having this makes it impossible to 'lock open' a permission.

contracts/src/registry/EnhancedAccessControl.sol
```diff
@@ -0,0 +1,302 @@
...
+
+    /**
+     * @dev Copies all roles from `srcAccount` in `srcResource` to `dstAccount` in `dstResource`.
+     *
+     * @param srcResource The resource to copy roles from.
+     * @param srcAccount The account to copy roles from.
+     * @param dstResource The resource to copy roles to.
+     * @param dstAccount The account to copy roles to.
+     */
+    function _copyRoles(bytes32 srcResource, address srcAccount, bytes32 dstResource, address dstAccount) internal virtual {
```
Arachnid, line 236: Does it make sense to copy roles across resources? I think this should probably be limited to copying them within a single resource.

Arachnid (changes requested):

contracts/src/registry/EnhancedAccessControl.sol
hiddentao: This isn't as straightforward since we need to check that the caller has the admin role for the given role - and we don't want to be looping through and checking the caller is an admin for every role passed in.

The alternative is to be able to set an admin for a bitmap of roles, but then that semantically doesn't make any sense.

Hence why granting/revoking one role at a time makes the most sense.

contracts/src/registry/EnhancedAccessControl.sol
hiddentao: This isn't as straightforward since we need to check that the caller has the admin role for the given role - and we don't want to be looping through and checking the caller is an admin for every role passed in.

The alternative is to be able to set an admin for a bitmap of roles, but then that semantically doesn't make any sense.

Hence why granting/revoking one role at a time makes the most sense.

contracts/src/registry/EnhancedAccessControl.sol
hiddentao: The issue is that the `DEFAULT_ADMIN_ROLE` can always revoke a role, thus I figured we needed an explicit locking mechanism to prevent that.

contracts/src/registry/BaseRegistry.sol
hiddentao: Was going to save that for a merge, but sure will add it in now

contracts/src/registry/EnhancedAccessControl.sol
hiddentao: I think it's possible that a role can be a role as well as an admin role - e.g registrar can be admin roles for setting the renewer admin role, even though registrars are themselves roles.

Arachnid: I don't think there's in harm in representing those with separate bits, though? If they're typically used together, the inheriting contract can simply grant or revoke both bits together.

contracts/src/registry/EnhancedAccessControl.sol
Arachnid: With the admin-bit-for-each-role I suggested above, there is no DEFAULT_ADMIN_ROLE - only individual admins for each role.

contracts/src/registry/EnhancedAccessControl.sol
Arachnid: Revoke all admin access over a role, without revoking the role itself for everyone - meaning it can still be done, but the set of accounts that can do it can't be changed any longer.

contracts/src/registry/EnhancedAccessControl.sol
```diff
@@ -0,0 +1,205 @@
...
+     *
+     * @param resource The resource to copy roles within.
+     * @param srcAccount The account to copy roles from.
+     * @param dstAccount The account to copy roles to.
+     */
+    function _copyRoles(bytes32 resource, address srcAccount, address dstAccount) internal virtual {
+        uint256 srcRoles = _roles[resource][srcAccount];
+        _roles[resource][dstAccount] |= srcRoles;
+        emit EACRolesCopied(resource, resource, srcAccount, dstAccount, srcRoles);
+    }
```
Arachnid, line 165: This also needs to mask out only the roles the user can grant, like above.

```suggestion
    function _copyRoles(address grantor, bytes32 resource, address srcAccount, address dstAccount) internal virtual {
        uint256 adminRoles = (_roles[resource][grantor] | _roles[ROOT_RESOURCE][grantor]) & 0xffffffffffffffffffffffffffffffff00000000000000000000000000000000;
        adminRoles |= adminRoles >> 128;
        uint256 srcRoles = _roles[resource][srcAccount] & adminRoles;
        _roles[resource][dstAccount] |= srcRoles;
        emit EACRolesCopied(resource, resource, srcAccount, dstAccount, srcRoles);
    }
```

contracts/src/registry/EnhancedAccessControl.sol
```diff
@@ -0,0 +1,205 @@
...
+ * NOTE:
+ * - Extending contracts must initialize their own default roles as this contract does not do so.
+ * - If two or more roles share the same admin role then these roles must ALWAYS be granted/revoked together.
+ */
+abstract contract EnhancedAccessControl is Context, ERC165 {
+    error EACUnauthorizedAccountRoles(bytes32 resource, uint256 roleBitmap, address account);
+    error EACUnauthorizedAccountAdminRoles(bytes32 resource, uint256 roleBitmap, address account);
+
+    event EACRolesGranted(bytes32 resource, uint256 roleBitmap, address account, address sender);
+    event EACRolesCopied(bytes32 srcResource, bytes32 dstResource, address account, address sender, uint256 roleBitmap);
```
Arachnid, line 33: I don't think we need separate events for 'granted' and 'copied'.

contracts/src/registry/EnhancedAccessControl.sol
```diff
@@ -0,0 +1,205 @@
...
+    }
+
+    /**
+     * @dev Copies all roles from `srcAccount` to `dstAccount` within the same resource.
+     *
+     * @param resource The resource to copy roles within.
+     * @param srcAccount The account to copy roles from.
+     * @param dstAccount The account to copy roles to.
+     */
+    function _copyRoles(bytes32 resource, address srcAccount, address dstAccount) internal virtual {
```
Arachnid, line 161: Perhaps there should be an easy way to *move* roles, too? We need to consider if we want to allow users to move roles to another account even if they aren't admin for those roles, as well.

contracts/src/registry/EnhancedAccessControl.sol
```diff
@@ -0,0 +1,205 @@
...
+        uint256 srcRoles = _roles[resource][srcAccount];
+        _roles[resource][dstAccount] |= srcRoles;
+        emit EACRolesCopied(resource, resource, srcAccount, dstAccount, srcRoles);
+    }
+
+    /**
+     * @dev Grants multiple roles to `account`.
+     */
+    function _grantRoles(bytes32 resource, uint256 roleBitmap, address account) internal virtual returns (bool){
```
Arachnid, line 170: As above, this needs to mask out only the bits the grantor can manage.

contracts/src/registry/EnhancedAccessControl.sol
```diff
@@ -0,0 +1,205 @@
...
+            return true;
+        } else {
+            return false;
+        }
+    }
+
+    /**
+     * @dev Attempts to revoke roles from `account` and returns a boolean indicating if roles were revoked.
+     */
+    function _revokeRoles(bytes32 resource, uint256 roleBitmap, address account) internal virtual returns (bool) {
```
Arachnid, line 186: As above - unless we also want to special case allowing a user to revoke their own roles without admin power.

contracts/src/registry/EnhancedAccessControl.sol
```diff
@@ -0,0 +1,205 @@
...
+        uint256 currentRoles = _roles[resource][account];
+        uint256 updatedRoles = currentRoles & ~roleBitmap;
+        
+        if (currentRoles != updatedRoles) {
+            _roles[resource][account] = updatedRoles;
+            emit EACRolesRevoked(resource, roleBitmap, account, _msgSender());
+            return true;
```
Arachnid, line 193: Elsewhere we give an error if they can't do the operation; we should be consistent here.

contracts/src/registry/EnhancedAccessControl.sol
hiddentao: Role copying is an internal-only function provided as a convenience utility. All public methods do have permission checks on them already.

contracts/src/registry/EnhancedAccessControl.sol
hiddentao: This is an internal method that gets called by the public `grant()` function which does have the permission checks. I want the internal methods to not be permissioned so that the subclasses have flexibility in terms of how they manage role assignments and revocations.

contracts/src/registry/EnhancedAccessControl.sol
hiddentao: The public method which calls this internal method already does throw an error. This is an internal method provided for convenience reasons in case concrete subclasses wish to operate differently.

contracts/src/registry/EnhancedAccessControl.sol
hiddentao: This would be the same as the `renounceRoles` method we had but then removed. Am going to leave it like this for now.

contracts/src/registry/EnhancedAccessControl.sol
hiddentao: This is an internal method that can be called by subclasses  and surrounded by whatever rules they want. To do a move a subclass could simply copy/grant followed by a revoke. I don't think we need to add anything more for this.

contracts/src/registry/EnhancedAccessControl.sol
```diff
@@ -0,0 +1,210 @@
...
+        } else {
+            return false;
+        }
+    }
+
+    /**
+     * @dev Revoke all roles for account within resource.
+     */
+    function _revokeAllRoles(bytes32 resource, address account) internal virtual returns (bool) {
+        return _revokeRoles(resource, _roles[resource][account], account);
```
Arachnid, line 194: This can avoid a storage read by just specifying `0xff..ff` for the `roleBitmap`.

contracts/src/registry/EnhancedAccessControl.sol
```diff
@@ -0,0 +1,210 @@
...
+     * 
+     * The settable roles are the roles that the account can grant/revoke.
+     * 
+     * @param resource The resource to get settable roles for.
+     * @param account The account to get settable roles for.
+     * @return The settable roles for `account` within `resource`.
+     */
+    function _getSettableRoles(bytes32 resource, address account) internal view virtual returns (uint256) {
+        uint256 adminRoleBitmap = (_roles[resource][account] | _roles[ROOT_RESOURCE][account]) & 0xffffffffffffffffffffffffffffffff00000000000000000000000000000000;
+        return adminRoleBitmap | adminRoleBitmap >> 128;
```
Arachnid, line 208: Just because order-of-operations makes me nervous here:

```suggestion
        return adminRoleBitmap | (adminRoleBitmap >> 128);
```

contracts/src/registry/NameWrapperRegistry.sol.bak
Arachnid, line 1: This needs renaming.

contracts/src/registry/Roles.sol
Arachnid, line 1: I think these should be set as constants on each implementing contract rather than as a global mapping.

contracts/src/registry/ETHRegistry.sol
```diff
@@ -1,26 +1,26 @@
...
 
-contract ETHRegistry is PermissionedRegistry, AccessControl, MetadataMixin, IETHRegistry {
-    bytes32 public constant REGISTRAR_ROLE = keccak256("REGISTRAR_ROLE");
 
+contract ETHRegistry is PermissionedRegistry, MetadataMixin, IETHRegistry {
     mapping(uint256 => address) public tokenObservers;
     
     constructor(IRegistryDatastore _datastore, IRegistryMetadata _metadata) PermissionedRegistry(_datastore) MetadataMixin(_metadata) {
-        _grantRole(DEFAULT_ADMIN_ROLE, msg.sender);
+        _grantRoles(ROOT_RESOURCE, ROLE_REGISTRAR | ROLE_REGISTRAR_ADMIN | ROLE_RENEW | ROLE_RENEW_ADMIN, _msgSender());
```
Arachnid, line 23: I like the simplicity of using only one access control mechanism for everything. But, mixing access control for our users and access control for system functions (registrar, renewal manager) makes me really nervous.

Granting a user one of these privileges on a non-root resource shouldn't matter, theoretically, but it seems entirely too close to being a vulnerability for comfort.

Thoughts?

contracts/src/registry/Roles.sol
hiddentao: I disagree. I think keeping it all in one ensures there are no accidental overlaps of roles, minimizing the chances of mistaken role assignments.

contracts/src/registry/ETHRegistry.sol
hiddentao: We could gatekeep things at higher level - e.g by overriding the grant/revoke methods - but I want to keep the lower levels (EAC) clean and pure. This is also why I wanted to put all the roles in the system into a single `Roles.sol` file so that there is no accidental overlap between roles.

contracts/src/registry/NameWrapperRegistry.sol.bak
hiddentao: Actually I needed to ask about this - the `UserRegistry` is supposed to be the name wrapper registry? Or are they two different things?

Arachnid: Ideally we equip the `UserRegistry` with all the features needed to be a replacement for the name wrapper. If that's not practical for some reason, we can have separate registries.

contracts/src/registry/Roles.sol
Arachnid: Each contract has its own set of roles. It's okay if two contracts use the same bits to mean different things.

contracts/src/registry/ETHRegistry.sol
Arachnid: I'm more wondering if we should use a different ACL mechanism (eg, OZ's) for access rights relating to the contract infrastructure. I have real worries about anything that combines permission checking for system-level privileges like controllers with permission checking for user-owned resources.

contracts/src/registry/EnhancedAccessControl.sol
```diff
@@ -25,6 +25,9 @@ abstract contract EnhancedAccessControl is Context, ERC165 {
     event EACRolesRevoked(bytes32 resource, uint256 roleBitmap, address account);
     event EACAllRolesRevoked(bytes32 resource, address account);
 
+    uint256 constant ALL_ROLES_BITMAP_MASK = 0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff;
```
Arachnid, line 28: I think just `ALL_ROLES` is clearer.

contracts/src/registry/EnhancedAccessControl.sol
```diff
@@ -25,6 +25,9 @@ abstract contract EnhancedAccessControl is Context, ERC165 {
...
+    uint256 constant ADMIN_ROLES_BITMAP_MASK = 0xffffffffffffffffffffffffffffffff00000000000000000000000000000000;
```
Arachnid, line 29: Likewise, `ADMIN_ROLES` is plenty clear.

contracts/src/registry/EnhancedAccessControl.sol
```diff
@@ -61,6 +74,15 @@ abstract contract EnhancedAccessControl is Context, ERC165 {
...
     }
 
+    /**
+     * @dev Constructor.
+     *
+     * @param _superAdmin The account to grant the `SUPERUSER` roles to.
+     */
+    constructor(address _superAdmin) {
+        _grantRoles(ROOT_RESOURCE, ROLE_SUPERUSER, _superAdmin);
+    }
```
Arachnid, line 84: I think how you had it previously was better - we should let implementers assign whatever default permissions they want in the constructor, rather than restricting them to granting superuser access to exactly one account.

contracts/src/registry/EnhancedAccessControl.sol
```diff
@@ -36,6 +39,16 @@ abstract contract EnhancedAccessControl is Context, ERC165 {
...
+    /**
+     * @dev The `SUPERUSER` role.
+     *
+     * This role automatically has all user roles and admin roles.
+     *
+     * Note:
+     * - The admin role bit for this role (128) is unnecessary, but still reserved for clarity purposes.
+     * - The SUPERUSER role is only effective in the `ROOT_RESOURCE`, not in subresources.
+     */
+    uint256 public constant ROLE_SUPERUSER = 1;
```
Arachnid, line 51: Why introduce this as something distinct from an account that just has all role bits set? It seems to add complexity but no new functionality.

Arachnid (changes requested):

contracts/src/registry/ETHRegistry.sol
hiddentao: Ok, let's finalize the acl system first and then think about it.

PR #20, opened 2025-02-26
feat:attempt to implement L1ETHRegistry for ejected names

mdtanrikulu:
Using @hiddentao 's ETHRegistry as a reference I tried to implement L1ETHRegistry. 

Note: Just now I see that the ticket is  assigned today already. So feel free to skip if there is another effort.

1.
> One design decision yet to be determined is whether renewal of ejected names
occurs on L2 or L1. If it occurs on L2, the L1 ejection controller is responsible for
updating the L1 .eth registry with new expiration dates propagated from L2.
Otherwise, a controller on the L1 .eth registry is responsible for renewing names
natively and sending a message to L2 to update the L2 .eth registry on the
change.

The code implements L1-native renewals, based on the feedback, but open to discussion.

  2. Not sure about `uri` function implementation, at the moment it returns emtpy string just as ETHRegistry, but I thought
  
  ```soliditiy
  function uri(uint256 tokenId) public view override returns (string memory) {
    (address resolver,) = datastore.getResolver(tokenId);
    return IResolver(resolver).uri(tokenId);
}
```

could be also an approach.

### DIFF
--- a/contracts/src/registry/L1ETHRegistry.sol
+++ b/contracts/src/registry/L1ETHRegistry.sol
@@ -18,27 +18,31 @@ import {PermissionedRegistry} from "./PermissionedRegistry.sol";
  * but receives names that have been ejected from L2.
  */
 contract L1ETHRegistry is PermissionedRegistry, AccessControl {
-    bytes32 public constant EJECTION_CONTROLLER_ROLE = keccak256("EJECTION_CONTROLLER_ROLE");
-
     error NameExpired(uint256 tokenId);
     error NameNotExpired(uint256 tokenId, uint64 expires);
     error CannotReduceExpiration(uint64 oldExpiration, uint64 newExpiration);
+    error OnlyEjectionController();
 
     event NameEjected(uint256 indexed tokenId, address owner, uint64 expires);
     event NameRenewed(uint256 indexed tokenId, uint64 newExpiration, address renewedBy);
     event NameMigratedToL2(uint256 indexed tokenId, address sendTo);
     event EjectionControllerChanged(address oldController, address newController);
 
-    address public ejectionController;
+    IL1EjectionController public ejectionController;
 
     constructor(IRegistryDatastore _datastore, address _ejectionController) PermissionedRegistry(_datastore) {
         _grantRole(DEFAULT_ADMIN_ROLE, msg.sender);
 
-        // Grant role to the ejection controller
+        // Set the ejection controller
         require(_ejectionController != address(0), "Ejection controller cannot be empty");
+        ejectionController = IL1EjectionController(_ejectionController);
+    }
 
-        ejectionController = _ejectionController;
-        _grantRole(EJECTION_CONTROLLER_ROLE, _ejectionController);
+    modifier onlyEjectionController() {
+        if (msg.sender != address(ejectionController)) {
+            revert OnlyEjectionController();
+        }
+        _;
     }
 
     /**
@@ -48,14 +52,10 @@ contract L1ETHRegistry is PermissionedRegistry, AccessControl {
     function setEjectionController(address _newEjectionController) external onlyRole(DEFAULT_ADMIN_ROLE) {
         require(_newEjectionController != address(0), "Ejection controller cannot be empty");
         
-        address oldController = ejectionController;
+        address oldController = address(ejectionController);
         
-        // Revoke role from the old controller
-        _revokeRole(EJECTION_CONTROLLER_ROLE, oldController);
-        
-        // Set the new controller and grant role
-        ejectionController = _newEjectionController;
-        _grantRole(EJECTION_CONTROLLER_ROLE, _newEjectionController);
+        // Set the new controller
+        ejectionController = IL1EjectionController(_newEjectionController);
         
         emit EjectionControllerChanged(oldController, _newEjectionController);
     }
@@ -91,7 +91,7 @@ contract L1ETHRegistry is PermissionedRegistry, AccessControl {
      */
     function ejectFromNamechain(uint256 labelHash, address owner, IRegistry registry, uint32 flags, uint64 expires)
         public
-        onlyRole(EJECTION_CONTROLLER_ROLE)
+        onlyEjectionController
         returns (uint256 tokenId)
     {
         tokenId = (labelHash & ~uint256(FLAGS_MASK)) | flags;
@@ -124,7 +124,7 @@ contract L1ETHRegistry is PermissionedRegistry, AccessControl {
      */
     function updateExpiration(uint256 tokenId, uint64 expires) 
         public 
-        onlyRole(EJECTION_CONTROLLER_ROLE)
+        onlyEjectionController
     {
         (address subregistry, uint96 flags) = datastore.getSubregistry(tokenId);
         uint64 oldExpiration = _extractExpiry(flags);
@@ -157,7 +157,7 @@ contract L1ETHRegistry is PermissionedRegistry, AccessControl {
         datastore.setSubregistry(tokenId, address(0), 0);
 
         // Notify the ejection controller to handle cross-chain messaging
-        IL1EjectionController(ejectionController).migrateToNamechain(tokenId, l2Owner, l2Subregistry, data);
+        ejectionController.migrateToNamechain(tokenId, l2Owner, l2Subregistry, data);
 
         emit NameMigratedToL2(tokenId, l2Owner);
     }
--- a/contracts/test/L1ETHRegistry.t.sol
+++ b/contracts/test/L1ETHRegistry.t.sol
@@ -24,28 +24,33 @@ contract TestL1ETHRegistry is Test, ERC1155Holder {
         datastore = new RegistryDatastore();
         ejectionController = new MockEjectionController();
         registry = new L1ETHRegistry(datastore, address(ejectionController));
-        registry.grantRole(registry.EJECTION_CONTROLLER_ROLE(), address(this));
     }
 
-    function test_eject_from_l2_unlocked() public {
+    function test_eject_from_namechain_unlocked() public {
         uint256 expectedId = (labelHash & 0xfffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff8);
 
+        vm.prank(address(ejectionController));
         uint256 tokenId = registry.ejectFromNamechain(labelHash, address(this), registry, 0, uint64(block.timestamp) + 86400);
+        
         vm.assertEq(tokenId, expectedId);
         vm.assertEq(registry.ownerOf(tokenId), address(this));
     }
 
-    function test_eject_from_l2_locked() public {
+    function test_eject_from_namechain_locked() public {
         uint32 flags = uint32(registry.FLAG_SUBREGISTRY_LOCKED() | registry.FLAG_RESOLVER_LOCKED());
         uint256 expectedId = (labelHash & 0xfffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff8) | flags;
 
+        vm.prank(address(ejectionController));
         uint256 tokenId =
             registry.ejectFromNamechain(labelHash, address(this), registry, flags, uint64(block.timestamp) + 86400);
+            
         vm.assertEq(tokenId, expectedId);
     }
 
-    function test_eject_from_l2_emits_events() public {
+    function test_eject_from_namechain_emits_events() public {
         vm.recordLogs();
+        
+        vm.prank(address(ejectionController));
         registry.ejectFromNamechain(labelHash, address(this), registry, 0, uint64(block.timestamp) + 86400);
 
         Vm.Log[] memory entries = vm.getRecordedLogs();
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by hiddentao** [comment]
Feel free to take the ticket!

File: contracts/src/registry/L1ETHRegistry.sol [hunk 1]
```diff
@@ -0,0 +1,244 @@
...
+    bytes32 public constant RENEWAL_CONTROLLER_ROLE = keccak256("RENEWAL_CONTROLLER_ROLE");
+
+    error NameExpired(uint256 tokenId);
+    error CannotReduceExpiration(uint64 oldExpiration, uint64 newExpiration);
+
+    event NameEjected(uint256 indexed tokenId, address owner, uint64 expires);
+    event NameRenewed(uint256 indexed tokenId, uint64 newExpiration, address renewedBy);
+    event NameRelinquished(uint256 indexed tokenId, address relinquishedBy);
+    event NameMigratedToL2(uint256 indexed tokenId, address sendTo);
+    event FallbackResolverSet(address resolver);
```
Arachnid, line 30: I don't think we need an event for this, since the registry will emit an event in any case.

contracts/src/registry/L1ETHRegistry.sol
```diff
@@ -0,0 +1,244 @@
...
+    event TokenObserverSet(uint256 indexed tokenId, address observer);
+
+    // Address of the fallback resolver for names not found in this registry
+    address public fallbackResolver;
```
Arachnid, line 34: This will be resolved via the parent registry's resolver field, no need to store it here.

contracts/src/registry/L1ETHRegistry.sol
```diff
@@ -0,0 +1,244 @@
...
+
+    // Map to track token observers for notification of renewal/ejection events
+    mapping(uint256 => address) public tokenObservers;
+
+    constructor(IRegistryDatastore _datastore) PermissionedRegistry(_datastore) {
+        _grantRole(DEFAULT_ADMIN_ROLE, msg.sender);
+    }
+
+    function uri(uint256 /*tokenId*/ ) public pure override returns (string memory) {
```
Arachnid, line 43: Best add a TODO here.

contracts/src/registry/L1ETHRegistry.sol
```diff
@@ -0,0 +1,244 @@
...
+    /**
+     * @dev Receive an ejected name from L2.
+     * @param label The label of the name
+     * @param owner The owner of the name
+     * @param registry The registry to use for the name
+     * @param flags Additional flags for the name
+     * @param expires Expiration timestamp
+     * @return tokenId The token ID of the ejected name
+     */
+    function ejectFromL2(string calldata label, address owner, IRegistry registry, uint96 flags, uint64 expires)
```
Arachnid, line 71: I think we should probably be using the label hash here, not the plaintext label.

contracts/src/registry/L1ETHRegistry.sol
```diff
@@ -0,0 +1,244 @@
...
+        public
+        onlyRole(EJECTION_CONTROLLER_ROLE)
+        returns (uint256 tokenId)
+    {
+        tokenId = (uint256(keccak256(bytes(label))) & ~uint256(FLAGS_MASK)) | flags;
+        flags = (flags & FLAGS_MASK) | (uint96(expires) << 32);
+
+        // If there is a previous owner, burn the token
+        address previousOwner = super.ownerOf(tokenId);
```
Arachnid, line 80: We need to check that the previous name is expired first, so that a malicious ejection controller can't override L1 names.

contracts/src/registry/L1ETHRegistry.sol
Arachnid, line 71: Either we should accept only a `uint32` of flags, or we should accept a combined flags+expiration as a single field. Accepting a `uint96` of which 64 bits have to be zero seems odd.

contracts/src/registry/L1ETHRegistry.sol
```diff
@@ -0,0 +1,244 @@
...
+    function setTokenObserver(uint256 tokenId, address _observer) external onlyTokenOwner(tokenId) {
+        tokenObservers[tokenId] = _observer;
+        emit TokenObserverSet(tokenId, _observer);
+    }
+
+    /**
+     * @dev Relinquish an ejected name
+     * @param tokenId The token ID of the name to relinquish
+     */
+    function relinquish(uint256 tokenId) external onlyTokenOwner(tokenId) {
```
Arachnid, line 130: How does this differ from `migrateToL2`?

contracts/src/registry/L1ETHRegistry.sol
```diff
@@ -0,0 +1,244 @@
...
+
+        emit NameRelinquished(tokenId, msg.sender);
+    }
+
+    /**
+     * @dev Migrate a name back to L2
+     * @param tokenId The token ID of the name to migrate
+     * @param sendTo The address to send the name to on L2
+     */
+    function migrateToL2(uint256 tokenId, address sendTo)
```
Arachnid, line 147: This needs to notify the ejection controller, so that control can be given back to the user on L2.

Arachnid (changes requested):

contracts/src/registry/L1ETHRegistry.sol
mdtanrikulu: Added TODO for now.

contracts/src/registry/L1ETHRegistry.sol
Arachnid: Why would we want to let people do this? It would seem it puts the name in a weird unrecoverable no-mans-land.

contracts/src/registry/L1ETHRegistry.sol
Arachnid: > So the parent (L1 .eth registry) registry here is the legacy and current ENSv1 L1 .eth registry?

I'm not quite sure what you mean. This contract is the `subregistry` for .eth. The .eth fallback resolver will be the `resolver` for .eth. There's no need for this contract to explicitly fall back to the resolver because the resolution process does that when no resolver for a more specific name is present.

The fallback resolver will look first on Namechain, and then fall back to looking up the name in the legacy registry.

mdtanrikulu: This phrase from docs made me confused;
`This contract is set as the resolver on the L1 .eth registry` 

Now clear.

contracts/src/registry/L1ETHRegistry.sol
mdtanrikulu: the idea was to give the flexibility of what L2ETH does can be done on ejected names too but, went through the migration design again, seemingly this complicates the coordination between L1 and namechain. user should migrate back to L2 for relinquish. removing it.

contracts/src/controller/IL1EjectionController.sol
```diff
@@ -0,0 +1,73 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.8.13;
+
+/**
+ * @title IL1EjectionController
+ * @dev Interface for the L1 ejection controller that facilitates migrations of names
+ * between L1 and L2, as well as handling renewals. The controller is responsible
+ * for cross-chain communication.
+ */
+interface IL1EjectionController {
```
Arachnid, line 10: In general, I don't think we can simply emit events and expect some indexing system to convert them into actions on an L2; instead we need to have a contract call the L1 side of a messaging bridge, which may use some kind of event system internally (but that's none of our concern). We can thus probably remove these events from the interface.

contracts/src/controller/IL1EjectionController.sol
```diff
@@ -0,0 +1,73 @@
...
+    /**
+     * @dev Called by the L1ETHRegistry when a user initiates a migration of a name
+     * back to L2. The controller is responsible for sending the cross-chain message
+     * to L2 to complete the migration.
+     *
+     * @param tokenId The token ID of the name being migrated
+     * @param l2Owner The address that will own the name on L2
+     * @param l2Subregistry The subregistry address to use on L2
+     */
+    function initiateL1ToL2Migration(uint256 tokenId, address l2Owner, address l2Subregistry) external;
```
Arachnid, line 44: Perhaps just `migrateToNamechain`? It should also take an extra `data` argument to allow the caller to trigger some operation on L2.

contracts/src/controller/IL1EjectionController.sol
```diff
@@ -0,0 +1,73 @@
...
+     * from L2 to L1. This function should verify the message source and then call
+     * the L1ETHRegistry to complete the ejection.
+     *
+     * @param labelHash The keccak256 hash of the label
+     * @param l1Owner The address that will own the name on L1
+     * @param l1Subregistry The subregistry address to use on L1
+     * @param flags The flags to set on the name
+     * @param expires The expiration timestamp of the name
+     */
+    function completeL2ToL1Ejection(
```
Arachnid, line 57: Maybe just `completeEjection`? And this should also take a `data` parameter to pass to the ERC1155 transfer function.

contracts/src/registry/L1ETHRegistry.sol
```diff
@@ -0,0 +1,235 @@
...
+    error NameNotExpired(uint256 tokenId, uint64 expires);
+    error CannotReduceExpiration(uint64 oldExpiration, uint64 newExpiration);
+
+    event NameEjected(uint256 indexed tokenId, address owner, uint64 expires);
+    event NameRenewed(uint256 indexed tokenId, uint64 newExpiration, address renewedBy);
+    event NameMigratedToL2(uint256 indexed tokenId, address sendTo);
+    event TokenObserverSet(uint256 indexed tokenId, address observer);
+
+    // Map to track token observers for notification of renewal events
+    mapping(uint256 => address) public tokenObservers;
```
Arachnid, line 33: I don't think we need this if the ejection controller is always notified, do we?

contracts/src/registry/L1ETHRegistry.sol
```diff
@@ -0,0 +1,235 @@
...
+    }
+
+    /**
+     * @dev Renew an ejected name.
+     * After renewal, the ejection controller should be notified to sync the update to L2.
+     *
+     * @param tokenId The token ID of the name to renew
+     * @param expires New expiration timestamp
+     */
+    function renew(uint256 tokenId, uint64 expires) public {
```
Arachnid, line 109: This lets users renew names indefinitely for free, which seems bad. We should add a controller role here and only allow that contract to do renewals.

contracts/src/registry/L1ETHRegistry.sol
```diff
@@ -0,0 +1,235 @@
...
+
+    event NameEjected(uint256 indexed tokenId, address owner, uint64 expires);
+    event NameRenewed(uint256 indexed tokenId, uint64 newExpiration, address renewedBy);
+    event NameMigratedToL2(uint256 indexed tokenId, address sendTo);
+    event TokenObserverSet(uint256 indexed tokenId, address observer);
+
+    // Map to track token observers for notification of renewal events
+    mapping(uint256 => address) public tokenObservers;
+
+    address public ejectionController;
```
Arachnid, line 35: We need a method for the owner to replace the ejection controller, and an event that's emitted when it changes.

Arachnid (changes requested):

contracts/src/registry/L1ETHRegistry.sol
```diff
@@ -0,0 +1,233 @@
...
+        
+        // Revoke role from the old controller
+        _revokeRole(EJECTION_CONTROLLER_ROLE, oldController);
+        
+        // Set the new controller and grant role
+        ejectionController = _newEjectionController;
+        _grantRole(EJECTION_CONTROLLER_ROLE, _newEjectionController);
+        
+        emit EjectionControllerChanged(oldController, _newEjectionController);
+    }
```
Arachnid, line 61: Given that there's only one ejection controller at a time, and we track it in a variable, we can eliminate the unnecessary use of the roles modifier here, and just directly check if the caller is the current ejection controller.

contracts/src/registry/L1ETHRegistry.sol
```diff
@@ -0,0 +1,233 @@
...
+    error NameExpired(uint256 tokenId);
+    error NameNotExpired(uint256 tokenId, uint64 expires);
+    error CannotReduceExpiration(uint64 oldExpiration, uint64 newExpiration);
+
+    event NameEjected(uint256 indexed tokenId, address owner, uint64 expires);
+    event NameRenewed(uint256 indexed tokenId, uint64 newExpiration, address renewedBy);
+    event NameMigratedToL2(uint256 indexed tokenId, address sendTo);
+    event EjectionControllerChanged(address oldController, address newController);
+
+    address public ejectionController;
```
Arachnid, line 32: This should be of type `IEjectionController`.

Arachnid (changes requested):

Arachnid (approved):

PR #23, opened 2025-03-03
implement upgradable UserRegistry

mdtanrikulu:


contracts/src/registry/ERC1155SingletonUpgradable.sol
```diff
@@ -0,0 +1,104 @@
...
+    function ownerOf(uint256 id) external view returns (address owner);
+}
+
+/**
+ * @title ERC1155SingletonUpgradeable
+ * @dev ERC1155 implementation that supports only a single token per ID.
+ * Stores owner instead of balance to allow fetching ownership information for a tokenId via `ownerOf`.
+ * This is an upgradeable version of the ERC1155Singleton contract.
+ */
+abstract contract ERC1155SingletonUpgradeable is
```
Arachnid, line 25: Can the shared code in this contract and `ERC1155Singleton` be refactored into an abstract base class inherited by both implementations?

contracts/src/registry/UserRegistry.sol
```diff
@@ -1,72 +1,285 @@
...
+    AccessControlUpgradeable, 
+    ERC1155SingletonUpgradeable,
+    IRegistry // IRegistry extends IERC1155Singleton which has ownerOf
+{
+    // =================== Constants ===================
+    
+    uint96 public constant FLAGS_MASK = 0x7;
+    uint96 public constant FLAG_SUBREGISTRY_LOCKED = 0x1;
+    uint96 public constant FLAG_RESOLVER_LOCKED = 0x2;
+    uint96 public constant FLAG_FLAGS_LOCKED = 0x4;
```
Arachnid, line 32: I think this is at odds with the work @hiddentao is doing to implement a bitmap-based roles model for permissions (see https://github.com/ensdomains/namechain/pull/19)

contracts/src/registry/UserRegistry.sol
Arachnid, line 1: It's really hard to tell what's going on here because of the linting changes that adjust the whole layout. Could you submit a new PR just doing the linting changes? We can then merge that and rebase this on top of it so we can see what's actually changing.

Arachnid (changes requested):

contracts/src/registry/UserRegistry.sol
mdtanrikulu: following the changes, will be updating and adapting when merged

hiddentao:
@mdtanrikulu Just merged in the new registry + roles architecture. UserRegistry was removed as it's code was merged into `PermissionedRegistry`. However there may still be a need for a derived `UserRegistry` contract which adds some additional helper methods and other logic on top. Are you still working on this ticket?

mdtanrikulu:
> @mdtanrikulu Just merged in the new registry + roles architecture. `UserRegistry` was removed as it's code was merged into `PermissionedRegistry`. However there may still be a need for a derived `UserRegistry` contract which adds some additional helper methods and other logic on top. Are you still working on this ticket?

hey @hiddentao yes, this ticket was in hold until your PR get merged. i will take a look on all changes to see what else needs to be covered from UserRegistry side. thanks!

hiddentao:
Great. I took another look at the new stuff and I think we can just re-use PermissionedRegistry for the user registry functionality. Not sure how that then ties in with the upgradeability - perhaps we still need a UserRegistry contract just for the sake of that.

mdtanrikulu:
> Great. I took another look at the new stuff and I think we can just re-use PermissionedRegistry for the user registry functionality. Not sure how that then ties in with the upgradeability - perhaps we still need a UserRegistry contract just for the sake of that.

I gave a try adding upgradability in PR 👉  #38. Since this PR is irrelevant with the newest design, closing it. We can continue the discussion on #38 👋

PR #29, opened 2025-03-26
feat: unified registry

hiddentao:
Related to #19


contracts/src/registry/PermissionedRegistry.sol
```diff
@@ -2,55 +2,264 @@
...
-    uint96 public constant FLAG_RESOLVER_LOCKED = 0x2;
-    uint96 public constant FLAG_FLAGS_LOCKED = 0x4;
+    uint256 public constant ROLE_SET_RESOLVER = 1 << 3;
+    uint256 public constant ROLE_SET_RESOLVER_ADMIN = ROLE_SET_RESOLVER << 128;
+
+    uint256 public constant ROLE_SET_FLAGS = 1 << 4;
+    uint256 public constant ROLE_SET_FLAGS_ADMIN = ROLE_SET_FLAGS << 128;
+
+    uint256 public constant MAX_EXPIRY = type(uint64).max;
+    uint96 public constant FLAGS_MASK = 0xffffffff; // 32 bits
```
Arachnid, line 37: I'm a bit concerned about the bytecode and runtime overhead of all these public constants. What do you think about making them private?

contracts/src/registry/PermissionedRegistry.sol
```diff
@@ -2,55 +2,264 @@
...
+        _mint(owner, tokenId, 1, "");
+        _grantRoles(tokenIdResource(tokenId), roleBitmap, owner);
+
+        datastore.setSubregistry(tokenId, address(registry), flags);
+        datastore.setResolver(tokenId, resolver, flags);
+
+        if (bytes(_uri).length > 0) {
+            metadataProvider.setTokenUri(tokenId, _uri);
+            emit URI(_uri, tokenId);
+        }
```
Arachnid, line 99: Aside from the security issue, I don't think we need per-token URIs in general. An interface that takes the token ID and generates it based on an algorithm should work fine.

contracts/src/registry/PermissionedRegistry.sol
```diff
@@ -2,55 +2,264 @@
...
+        
+        emit NewSubname(tokenId, label);
+
+        return tokenId;
+    }
+
+    function setTokenObserver(uint256 tokenId, address _observer) external onlyTokenOwner(tokenId) {
```
Arachnid, line 106: Should there be a permission to set a token observer?

contracts/src/registry/PermissionedRegistry.sol
```diff
@@ -2,55 +2,264 @@
...
+        tokenObservers[tokenId] = _observer;
+        emit TokenObserverSet(tokenId, _observer);
+    }
+
+    function renew(uint256 tokenId, uint64 expires) public onlyRootRoles(ROLE_RENEW) {
```
Arachnid, line 111: Perhaps this should be `onlyRoles` instead of `onlyRootRoles`; it could be useful to grant a user the right to renew their own name (cost-free, for user registries for example). We'd need to mask this bit out in the .eth registrar.

contracts/src/registry/PermissionedRegistry.sol
```diff
@@ -2,55 +2,264 @@
...
+            newTokenId = (tokenId & ~uint256(FLAGS_MASK)) | (newFlags & FLAGS_MASK);
+            if (tokenId != newTokenId) {
+                address owner = ownerOf(tokenId);
+                _mint(owner, newTokenId, 1, "");
+                _burn(owner, tokenId, 1);
+            }
+        } else {
+            newTokenId = tokenId;
+        }
+    }
```
Arachnid, line 216: Are we 100% sure we still need flags, now that we have the new permission model?

contracts/src/registry/PermissionedRegistry.sol
Arachnid, line 1: There's a vulnerability here whereby a name that expires and is reregistered still has all the old name's ACLs. Likewise, we need to provide a way to invalidate ACLs when a name is transferred to a new owner.

One way to do this is with a 'version' field, but it's an extra element to store. We could mask off bits in the token ID to store this. Thoughts?

Arachnid (changes requested):

contracts/src/registry/PermissionedRegistry.sol
hiddentao: That's fine, but we then have to accept front-end brittleness, i.e things could go wrong.

contracts/src/registry/PermissionedRegistry.sol
hiddentao: Well, only token owner can set it. That should do, right?

contracts/src/registry/PermissionedRegistry.sol
hiddentao: Sounds like we need a "Revoke all roles for all users" thing, like I mentioned before :)

contracts/src/registry/PermissionedRegistry.sol
hiddentao: Well we have support for any scheme. That's the point the Registry metadata implementations. Are you saying to assume we'll never externally set the uri per tokenId?

contracts/src/registry/PermissionedRegistry.sol
hiddentao: Well, I wasn't sure what else you guys wanted flags for so left them in...should I remove them? This would also mean being able to cleanup the "token id changes when flags change" logic, right?

Arachnid: If we're not using them now, then yes we can.

The issue the token ID change was created to fix is present here in another form - we don't want a situation where someone can grant themselves permissions just prior to transferring the domain to the new owner, and we need a way to make sure the user has 'clean title' to the name they just bought. Needs some thought on how to handle this.

contracts/src/registry/PermissionedRegistry.sol
Arachnid: Right, I'm suggesting it would be more flexible if that's a right that can be delegated like others here.

contracts/src/registry/PermissionedRegistry.sol
Arachnid: I can't see a need to set per-token URIs in the root, .eth, or user registries. So I think for now we can leave them out. A subclass of this could always introduce them if needed.

contracts/src/registry/PermissionedRegistry.sol
Arachnid: Right, but that needs to be compatible with also being able to permanently revoke permissions on a name. Any thoughts on how to achieve both objectives?

hiddentao:
@Arachnid 

i've removed flags completely. Note that `RegistryDatastore` now just stores the expiry and address in the `uint256`, leaving the leftover bits doing nothing for now.

As for solving the two issues you mentioned (people assigning roles to themselves prior to a transfer / expired names still have old role assignments when being registered again), i've implemented a versioning system in the tokenIdResource() method, so it's the `PermissionedRegistry` level. A benefit of this is that the NFT token id now doesn't need to change, bringing stability across the front-end side of things - i.e. once a name is created the NFT token id for it never changes.

contracts/src/registry/PermissionedRegistry.sol
hiddentao: See https://github.com/ensdomains/namechain/pull/29#issuecomment-2760497409

contracts/src/registry/PermissionedRegistry.sol
hiddentao: See https://github.com/ensdomains/namechain/pull/29#issuecomment-2760497409

contracts/src/registry/PermissionedRegistry.sol
```diff
@@ -2,55 +2,215 @@
...
+        for (uint256 i = 0; i < ids.length; i++) {
+            tokenIdResourceVersion[ids[i]]++;
+        }
     }
 }
+
+/**
+ * @dev Observer pattern for events on existing tokens.
+ */
+interface PermissionedRegistryTokenObserver {
```
Arachnid, line 213: Can probably just call this `TokenObserver`.

Arachnid (changes requested):

contracts/src/registry/BaseUriRegistryMetadata.sol
```diff
@@ -22,6 +23,6 @@ contract BaseUriRegistryMetadata is AccessControl, IRegistryMetadata {
     }
 
     function supportsInterface(bytes4 interfaceId) public view override returns (bool) {
-        return interfaceId == type(IRegistryMetadata).interfaceId || super.supportsInterface(interfaceId);
+        return interfaceId == type(IRegistryMetadata).interfaceId || EnhancedAccessControl.supportsInterface(interfaceId);
```
Arachnid, line 26: The original implementation here is correct.

contracts/src/registry/ETHRegistrar.sol
```diff
@@ -22,16 +27,17 @@ contract ETHRegistrar is IETHRegistrar, AccessControl {
     error NameNotAvailable(string name);
     error InsufficientValue(uint256 required, uint256 provided);
 
-    IETHRegistry public immutable registry;
+    IPermissionedRegistry public immutable registry;
```
Arachnid, line 30: We should probably define an interface specific to a registry-with-register-and-renew functions that's more general than IPermissionedRegistry.

contracts/src/registry/EnhancedAccessControl.sol
```diff
@@ -188,23 +188,24 @@ abstract contract EnhancedAccessControl is Context, ERC165 {
...
         uint256 srcRoles = _roles[resource][srcAccount];
-        _grantRoles(resource, srcRoles, dstAccount);
+        _grantRoles(resource, srcRoles, dstAccount, enableCallbacks);
     }
 
     /**
      * @dev Grants multiple roles to `account`.
      */
-    function _grantRoles(bytes32 resource, uint256 roleBitmap, address account) internal virtual returns (bool) {
+    function _grantRoles(bytes32 resource, uint256 roleBitmap, address account, bool enableCallbacks) internal virtual returns (bool) {
```
Arachnid, line 199: Perhaps `executeCallbacks` instead of `enableCallbacks`; 'enable' implies a persistent change.

contracts/src/registry/PermissionedRegistry.sol
```diff
@@ -2,55 +2,269 @@
...
+        }
+        _;
     }
 
-    function _setFlags(uint256 tokenId, uint96 _flags)
-        internal
-        withSubregistryFlags(tokenId, FLAG_FLAGS_LOCKED, 0)
-        returns(uint96 newFlags)
+    constructor(IRegistryDatastore _datastore, IRegistryMetadata _metadata) BaseRegistry(_datastore) MetadataMixin(_metadata) {
+        _grantRoles(ROOT_RESOURCE, ALL_ROLES, _msgSender(), false);
```
Arachnid, line 48: When instantiating this in a trustless context - for example, the .eth registry - we need to revoke or not set `ROLE_SET_SUBREGISTRY` and `ROLE_SET_RESOLVER` and `ROLE_SET_TOKEN_OBSERVER` for the contract owner. We could do this by calling the relevant method after deployment, but perhaps it would be cleaner to add a `deployerRoles` argument to the constructor?

contracts/src/registry/PermissionedRegistry.sol
```diff
@@ -2,55 +2,269 @@
...
+    }
+
+    function register(string calldata label, address owner, IRegistry registry, address resolver, uint256 roleBitmap, uint64 expires)
+        public
+        onlyRootRoles(ROLE_REGISTRAR)
+        returns (uint256 tokenId)
+    {
+        tokenId = NameUtils.labelToTokenId(label);
+
+        (, uint64 oldExpiry, uint32 tokenIdVersion) = datastore.getSubregistry(tokenId);
```
Arachnid, line 80: Perhaps this should be combined with line `90` using a utility function that returns `(versionedTokenId, expiry)`?

contracts/src/registry/PermissionedRegistry.sol
```diff
@@ -2,55 +2,269 @@
...
+            revert CannotSetPastExpiration(expires);
+        }
+
+        tokenId = _constructVersionedTokenId(tokenId, tokenIdVersion);
+
+        // if there is a previous owner, burn the token
+        address previousOwner = super.ownerOf(tokenId);
+        if (previousOwner != address(0)) {
+            _burn(previousOwner, tokenId, 1);
+            tokenId = _regenerateTokenId(tokenId); // so we have a fresh acl
```
Arachnid, line 96: We already know the `registry` and `expires` here; can we pass them in to this function to avoid re-fetching them?

contracts/src/registry/PermissionedRegistry.sol
```diff
@@ -2,55 +2,269 @@
...
         }
+        return super.ownerOf(tokenId);
+    }
+
+    function register(string calldata label, address owner, IRegistry registry, address resolver, uint256 roleBitmap, uint64 expires)
+        public
+        onlyRootRoles(ROLE_REGISTRAR)
+        returns (uint256 tokenId)
+    {
+        tokenId = NameUtils.labelToTokenId(label);
```
Arachnid, line 78: I'm concerned about passing around multiple different things called `tokenId` that have different values - versioned and unversioned. Could we either rename unversioned uses, or avoid having this intermediate value in a variable?

contracts/src/registry/RegistryDatastore.sol
```diff
@@ -2,43 +2,47 @@
 pragma solidity >=0.8.13;
 
 import {IRegistryDatastore} from "./IRegistryDatastore.sol";
+import {NameUtils} from "../utils/NameUtils.sol";
 
 contract RegistryDatastore is IRegistryDatastore {
-    uint256 LABEL_HASH_MASK = 0xffffffffffffffffffffffffffffffffffffffffffffffffffffffff00000000;
```
Arachnid, line 7: I think we still need to mask out the lower bits in the datastore, because that provides a guarantee to anyone reading the datastore storage about how token IDs will be treated.

Arachnid (changes requested):

contracts/src/registry/RegistryDatastore.sol
hiddentao: It still masks it out - I've just abstracted the logic into `NameUtils.getCanonicalId()`

contracts/src/registry/PermissionedRegistry.sol
hiddentao: I'll use `tokenId` and `canonicalId` where the latter means with the bits zeroed out.

contracts/src/registry/ETHRegistrar.sol
hiddentao: I need more guidance here. Should it have relinquish as well? what about the getters to fetch expiry, etc? and that setters for subregistry and resolver value?

Arachnid: Ideally it should be a useful interface that's a superset of `IRegistry`, but not specific to `PermissionedRegistry`. So it would need to include `register`, `renew` and any other methods this contract uses, but needn't include methods this contract doesn't use unless they're naturally part of the same interface. So the getters for expiry etc would naturally be included because their setters are, for example.

hiddentao: Ok, I've just updated `IRegistry` directly

contracts/src/registry/IStandardRegistry.sol
```diff
@@ -0,0 +1,104 @@
...
+     */
+    function getTokenIdResource(uint256 tokenId) external view returns(bytes32);
+
+    /**
+     * @dev Get the token ID of a resource.
+     * @param resource The resource to fetch the token ID for.
+     * @return The token ID of the resource.
+     */
+    function getResourceTokenId(bytes32 resource) external view returns (uint256);
+}
```
Arachnid, line 104: Do these need to be part of the public interface?

contracts/src/registry/PermissionedRegistry.sol
```diff
@@ -2,55 +2,295 @@
...
+
+    function register(string calldata label, address owner, IRegistry registry, address resolver, uint256 roleBitmap, uint64 expires)
+        public
+        override
+        onlyRootRoles(ROLE_REGISTRAR)
+        returns (uint256 tokenId)
+    {
+        uint64 oldExpiry;
+        uint32 tokenIdVersion;
+        (tokenId, oldExpiry, tokenIdVersion) = this.getNameData(label);
```
Arachnid, line 81: Calling via `this` makes it an external call, which is less efficient. Is this needed here?

contracts/src/registry/PermissionedRegistry.sol
```diff
@@ -2,55 +2,295 @@
...
+        }
+
+        // if there is a previous owner, burn the token
+        address previousOwner = super.ownerOf(tokenId);
+        if (previousOwner != address(0)) {
+            _burn(previousOwner, tokenId, 1);
+            tokenId = _regenerateTokenId(tokenId, address(registry), expires, tokenIdVersion); // so we have a fresh acl
+        } else {
+            datastore.setSubregistry(tokenId, address(registry), expires, tokenIdVersion);
+        }
```
Arachnid, line 98: If you change `_regenerateTokenId` to take the new version instead of the old one, you can remove the `else` clause, increment the `tokenIdVersion` in the `if` clause, and call `_regenerateTokenId` outside the if clause to handle both cases. Maybe call it `generateTokenId` instead?

```suggestion
        if (previousOwner != address(0)) {
            _burn(previousOwner, tokenId, 1);
            tokenIdVersion++; // so we have a fresh acl
        }
        tokenId = _generateTokenId(tokenId, address(registry), expires, tokenIdVersion); 
```

contracts/src/registry/PermissionedRegistry.sol
```diff
@@ -2,55 +2,295 @@
...
+
+    /**
+     * @dev Override the base registry _update function to transfer the roles to the new owner when the token is transferred.
+     */
+    function _update(address from, address to, uint256[] memory ids, uint256[] memory values) internal virtual override {
+        super._update(from, to, ids, values);
+
+        for (uint256 i = 0; i < ids.length; ++i) {
+            /*
+            in _regenerateToken, we burn the token and then mint a new one. This flow below ensures the roles go from owner => zeroAddr => owner
```
Arachnid, line 222: It's not clear to me what this comment means. Isn't the order the reverse of what's implied?

contracts/src/registry/PermissionedRegistry.sol
```diff
@@ -2,55 +2,295 @@
...
+     */
+    function _constructTokenId(uint256 id, uint32 tokenIdVersion) internal pure returns (uint256 newTokenId) {
+        newTokenId = NameUtils.getCanonicalId(id) | tokenIdVersion;
     }
 }
+
+/**
+ * @dev Observer pattern for events on existing tokens.
+ */
+interface TokenObserver {
```
Arachnid, line 293: This should be in its own file, so that anyone wanting to implement an observer doesn't need to include the entire registry source code.

contracts/src/registry/ETHRegistrar.sol
```diff
@@ -142,24 +142,20 @@ contract ETHRegistrar is IETHRegistrar, AccessControl {
...
-            revert InsufficientValue(totalPrice, msg.value);
-        }
+        uint256 totalPrice = checkPrice(name, duration);
 
-        _consumeCommitment(name, duration, makeCommitment(name, owner, secret, address(subregistry), resolver, flags, duration));
+        _consumeCommitment(name, duration, makeCommitment(name, owner, secret, address(subregistry), resolver, duration));
 
-        tokenId = registry.register(name, owner, subregistry, resolver, flags, uint64(block.timestamp) + duration);
+        uint64 expiry = uint64(block.timestamp) + duration;
+        tokenId = registry.register(name, owner, subregistry, resolver, ALL_ROLES, expiry);
```
Arachnid, line 152: This needs adjusting - as it stands it will allow the user to renew their own name directly. Either this should mask out `ROLE_RENEW` and `ROLE_RENEW_ADMIN` or it should be set to only the safe roles.

contracts/src/registry/PermissionedRegistry.sol
hiddentao: the flow is:

1. burn
2. mint

in (1) the roles are copied from owner to zero address, in (2) they are copied from zero address back to the owner.

Arachnid (approved):

Arachnid (approved):

PR #30, opened 2025-03-26
refactor: create L1, L2, common folders

hiddentao:


makoto:
L1/L1EjectionController can just be called as L1/EjectionController

Arachnid (approved):

hiddentao:
@makoto I think filenames should match the contract names to avoid any confusion, so will keep them as they are.

PR #33, opened 2025-04-01
feat: callbacks for role grants and revokes

hiddentao:


contracts/src/registry/EnhancedAccessControl.sol
```diff
@@ -6,6 +6,8 @@ pragma solidity ^0.8.20;
 import {Context} from "@openzeppelin/contracts/utils/Context.sol";
 import {ERC165} from "@openzeppelin/contracts/utils/introspection/ERC165.sol";
 
+
+
```
Arachnid, line 10: Can you remove the excess whitespace?

PR #36, opened 2025-04-07
Feature/bet 348 create devnet command

mdtanrikulu:


Arachnid:
This is a good start, but I think we need a modular deployment system like we have in ens-contracts. Is there any existing equivalent for Forge?

adraffy:
What does "modular deployment system" mean?

I think we should just use [blocksmith.js](https://github.com/adraffy/blocksmith.js) as `setup.ts` and `runDevnet.ts` can be replaced with a few lines of code.  It also makes it significantly easier to build external contracts and write tests.  It's just a programmatic wrapper around `forge` and `anvil`.

adraffy:
Also, if blocksmith is too experimental, [tevm](https://www.tevm.sh/) might be the way to go.

Arachnid:
> What does "modular deployment system" mean?

Something that allows for deploying individual contracts as needed, and saving the deployment state. We want to be able to add new deployment units and when we deploy new or modified contracts, only deploy those contracts that have changed.

PR #38, opened 2025-04-10
implement new upgradable User Registry with PermissionedRegistry inheritence

mdtanrikulu:


contracts/src/common/RegistryRolesMixin.sol
```diff
@@ -17,4 +17,7 @@ abstract contract RegistryRolesMixin {
 
     uint256 internal constant ROLE_SET_TOKEN_OBSERVER = 1 << 4;
     uint256 internal constant ROLE_SET_TOKEN_OBSERVER_ADMIN = ROLE_SET_TOKEN_OBSERVER << 128;
+
+    uint256 internal constant ROLE_UPGRADE = 1 << 5;
```
hiddentao, line 21: If this role is only being used in one contract (`UserRegistry`) then it doesn't need to go into this file - it should stay within `UserRegistry` instead.

contracts/test/UserRegistry.t.sol
```diff
@@ -0,0 +1,277 @@
...
+        assertEq(upgradedProxy.version(), 2, "Version should be 2 after upgrade");
+    }
+    
+    function test_Revert_unauthorized_upgrade() public {
+        // Deploy a new implementation
+        UserRegistryV2Mock newImplementation = new UserRegistryV2Mock();
+        
+        // User1 tries to upgrade without permission
+        vm.prank(user1);
+        vm.expectRevert(); // Should revert because user1 doesn't have upgrade rights
```
hiddentao, line 230: This should test for the exact reversion error so that we know it's the revert we expect.

contracts/test/UserRegistry.t.sol
```diff
@@ -0,0 +1,277 @@
...
+        );
+        
+        // Verify registration was successful
+        assertEq(proxy.ownerOf(tokenId), user2, "Domain should be owned by user2");
+    }
+    
+    function test_Revert_unauthorized_registration() public {
+        // User1 tries to register a domain without ROLE_REGISTRAR
+        vm.prank(user1);
+        vm.expectRevert(); // Should revert because user1 doesn't have ROLE_REGISTRAR
```
hiddentao, line 169: This should test for the exact reversion error so that we know it's the revert we expect.

contracts/test/UserRegistry.t.sol
```diff
@@ -0,0 +1,277 @@
...
+            address(0),
+            ROLE_SET_SUBREGISTRY | ROLE_SET_RESOLVER,
+            uint64(block.timestamp + 365 days)
+        );
+    }
+    
+    function test_Revert_unauthorized_role_grant() public {
+        // User1 tries to grant roles without permission
+        vm.prank(user1);
+        vm.expectRevert(); // Should revert because user1 doesn't have ROLE_REGISTRAR_ADMIN
```
hiddentao, line 183: This should test for the exact reversion error so that we know it's the revert we expect.

hiddentao (changes requested):

contracts/src/common/RegistryRolesMixin.sol
mdtanrikulu: this role could actually be available for any upgradable  logic as a future proof but I am alright moving it back to UserRegistry at this stage.

contracts/src/L2/UserRegistry.sol
```diff
@@ -0,0 +1,69 @@
...
+        address admin
+    ) public initializer {
+        require(admin != address(0), "Admin cannot be zero address");
+        
+        // Initialize datastore
+        datastore = _datastore;
+        
+        // Initialize metadata provider
+        if (address(_metadata) == address(0)) {
+            // Create a new SimpleRegistryMetadata if none is provided
```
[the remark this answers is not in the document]
mdtanrikulu: or we can also accept this as only constructor parameter for UserRegistry instead of initializer parameter then no need to duplicate the logic I guess.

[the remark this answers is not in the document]
Arachnid: The PermissionedRegistry constructor won't be called because it's being cloned via proxy, right?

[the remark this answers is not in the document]
mdtanrikulu: oh right, since the storage layout will be on proxy, that won't work. then what @hiddentao says seem like the only solution.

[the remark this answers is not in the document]
hiddentao: @Arachnid Yeah, but the implementation contractor constructor (and it's base class constructors) still get executed, even if the contract itself is then only called later via the proxy. Which means it's still going to wastefully deploy a `SimpleRegistryMetadata` instance, as it currently stands. Is this not so?

hiddentao:
@mdtanrikulu All looks good except for the registry metadata instance creation issue that we're still discussing.

Arachnid:
I'm fine with the implementation contract deploying a redundant metadata service that doesn't get used - it's only one instance.

Arachnid (approved):

PR #39, opened 2025-04-15
feat: L2 ejection controller

hiddentao:


hiddentao:
Notes:

- I'm not sure if the events related to ejection should be emitted by just `L2EjectionController` or also by `L2ETHRegistry`.
- The `L2ETHRegistry` class just adds a single method, and am pondering whether I can get rid of it altogether and place the method into `L2EjectionController` - the required token observer setting + transfer approval could be set when the name gets registered in the first place - any feedback on this welcome

contracts/src/L2/L2EjectionController.sol
```diff
@@ -0,0 +1,106 @@
...
+    function ejectToL1(uint256 tokenId, address l1Owner, address l1Subregistry) external {
+        if (registry.ownerOf(tokenId) != address(this)) {
+            revert NotTokenOwner(tokenId);
+        }
+
+        uint64 expiry = registry.getExpiry(tokenId);
+
+        registry.setSubregistry(tokenId, IRegistry(address(0)));
+
+        // bridge will listen for this event
```
makoto, line 43: Shouldn't this be calling Linea's `L2MessageService.sendMessage` according to their doc https://docs.linea.build/technology/message-service ?

hiddentao: As discussed, we'll probably have a pluggable bridge system, leaving the ejection controller itself to not have to worry about this.

contracts/src/L1/L1ETHRegistry.sol
```diff
@@ -7,55 +7,46 @@ import {IERC1155Singleton} from "../common/IERC1155Singleton.sol";
...
-        
+    function setEjectionController(IL1EjectionController _newEjectionController) external onlyRoles(ROOT_RESOURCE, ROLE_SET_EJECTION_CONTROLLER) {
         address oldController = address(ejectionController);
-        
-        // Set the new controller
-        ejectionController = IL1EjectionController(_newEjectionController);
-        
-        emit EjectionControllerChanged(oldController, _newEjectionController);
+        _setEjectionController(_newEjectionController);
+        emit EjectionControllerChanged(oldController, address(_newEjectionController));
```
Arachnid, line 49: Shouldn't this be emitted in `_setEjectionController`?

contracts/src/L2/IL2EjectionController.sol
Arachnid, line 1: If we name the methods a bit more generically - like `eject` and `receive`, do we actually need two different interfaces?

contracts/src/L2/L2ETHRegistry.sol
```diff
@@ -0,0 +1,66 @@
...
+
+/**
+ * @title L2ETHRegistry
+ * @dev L2 contract for .eth that holds .eth names.
+ */
+contract L2ETHRegistry is PermissionedRegistry, EjectionControllerMixin {
+    error NameNotExpired(uint256 tokenId, uint64 expires);
+
+    event NameEjected(uint256 indexed tokenId, address owner, uint64 expires);
+    event NameMigratedToL2(uint256 indexed tokenId, address sendTo);
```
Arachnid, line 24: This isn't emitted anywhere.

contracts/src/L2/L2ETHRegistry.sol
```diff
@@ -0,0 +1,66 @@
...
+     * @param tokenId The token ID of the name
+     * @param l1Owner The new owner of the name on L1
+     * @param l1Subregistry The new subregistry to use for the name on L1
+     * @param l1Resolver The new resolver to use for the name on L1
+     */
+    function eject(uint256 tokenId, address l1Owner, address l1Subregistry, address l1Resolver)
+        public
+        onlyTokenOwner(tokenId)
+    {
+        setTokenObserver(tokenId, ejectionController);
```
Arachnid, line 53: I think the ejection controller should probably do this in its on-received handler for the token.

In fact, I think we can eliminate the explicit method on the ejection controller, and just pass the ejection data in the data field of the transfer.

contracts/src/L2/IL2EjectionController.sol
```diff
@@ -0,0 +1,38 @@
...
+
+    /**
+     * @dev Called by the cross-chain messaging system when a name is being migrated back to L2.
+     *
+     * @param labelHash The keccak256 hash of the label
+     * @param l2Owner The address that will own the name on L2
+     * @param l2Subregistry The subregistry address to use on L2
+     * @param l2Resolver The resolver address to use on L2
+     */
+    function completeMigrationToL2(
```
Arachnid, line 32: I don't think this needs to be part of the interface - it can be different for each ejection controller implementation.

contracts/src/L2/L2EjectionController.sol
```diff
@@ -0,0 +1,110 @@
...
+        if (registry.ownerOf(tokenId) != address(this)) {
+            revert NotTokenOwner(tokenId);
+        }
+
+        registry.setSubregistry(tokenId, IRegistry(l2Subregistry));
+        registry.setResolver(tokenId, l2Resolver);
+        registry.safeTransferFrom(address(this), l2Owner, tokenId, 1, "");
+
+        emit NameMigratedToL2(tokenId, l2Owner, l2Subregistry, l2Resolver);
+    }
```
Arachnid, line 71: I think that we should probably have subclasses of the ejection controller; this would be the implementation for the dummy one, with a different one for linea.

contracts/src/L2/L2EjectionController.sol
```diff
@@ -0,0 +1,110 @@
...
+        return this.onERC1155BatchReceived.selector;
+    }
+
+    /**
+     * Implements ITokenObserver.onRenew
+     */
+    function onRenew(uint256 tokenId, uint64 expires, address renewedBy) external {
+        if (registry.ownerOf(tokenId) == address(this)) {
+            // this will get picked up by the bridge
+            emit NameRenewed(tokenId, expires, renewedBy);
```
Arachnid, line 100: Likewise here - this should be in a subclass.

contracts/src/common/EjectionControllerMixin.sol
```diff
@@ -0,0 +1,10 @@
+// SPDX-License-Identifier: MIT
```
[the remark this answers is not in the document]
hiddentao: Have now replaced this with a new base class - `ETHRegistry`

contracts/src/L2/IL2EjectionController.sol
hiddentao: Not sure which two interfaces you meant, but I've done some refactoring now to change how/where thing are defined.

contracts/src/L2/L2ETHRegistry.sol
hiddentao: Done

contracts/src/L2/L2EjectionController.sol
hiddentao: But don't we want this base class to atleast emit an event?

contracts/test/L1ETHRegistry.t.sol
```diff
@@ -40,7 +41,7 @@ contract TestL1ETHRegistry is Test, ERC1155Holder, RegistryRolesMixin, EnhancedA
         datastore = new RegistryDatastore();
         ejectionController = new MockEjectionController();
         registryMetadata = new MockRegistryMetadata();
-        registry = new L1ETHRegistry(datastore, registryMetadata, address(ejectionController));
+        registry = new L1ETHRegistry(datastore, registryMetadata, IL1EjectionController(address(ejectionController)));
```
mdtanrikulu, line 44: i think you don't need this conversion, `MockEjectionController` is already implements IL1EjectionController, passing `ejectionController` should be fine.

contracts/test/L1ETHRegistry.t.sol
```diff
@@ -10,110 +10,118 @@ import "../src/L1/L1ETHRegistry.sol";
...
         ejectionController = new MockEjectionController();
         registryMetadata = new MockRegistryMetadata();
-        registry = new L1ETHRegistry(datastore, address(ejectionController), registryMetadata);
+        registry = new L1ETHRegistry(datastore, registryMetadata, ejectionController);
     }
 
     function test_eject_from_namechain_unlocked() public {
         vm.prank(address(ejectionController));
-        uint256 tokenId = registry.ejectFromNamechain(labelHash, address(this), registry, uint64(block.timestamp) + 86400);
+        uint256 tokenId = registry.ejectFromNamechain(labelHash, address(this), IRegistry(address(registry)), MOCK_RESOLVER, uint64(block.timestamp) + 86400);
```
mdtanrikulu, line 48: same redundant conversion applies here and 8-9 other lines in the same file as well for `IRegistry(address(registry))` since `L1ETHRegistry` implements `IRegistry` in the base.

contracts/src/common/EjectionController.sol
```diff
@@ -0,0 +1,92 @@
...
+
+    IPermissionedRegistry public immutable registry;
+
+    struct TransferData {
+        string label;
+        address newOwner;
+        address newSubregistry;
+        address newResolver;
+        uint256 newRoleBitmap;
+        uint64 newExpires;
```
Arachnid, line 23:
```suggestion
        address owner;
        address subregistry;
        address resolver;
        uint256 roleBitmap;
        uint64 expires;
```

contracts/src/common/EjectionController.sol
```diff
@@ -0,0 +1,92 @@
...
+
+    constructor(IPermissionedRegistry _registry) {
+        registry = _registry;
+    }
+
+    /**
+     * Implements ERC165.supportsInterface
+     */
+    function supportsInterface(bytes4 interfaceId) external pure returns (bool) {
+        return interfaceId == type(EjectionController).interfaceId || interfaceId == type(IERC1155Receiver).interfaceId;
```
Arachnid, line 34: You need to call `super` here as well.

contracts/src/L2/L2EjectionController.sol
```diff
@@ -0,0 +1,64 @@
...
+        uint256 tokenId,
+        TransferData memory transferData
+    ) internal virtual {
+        if (registry.ownerOf(tokenId) != address(this)) {
+            revert NotTokenOwner(tokenId);
+        }
+
+        registry.setSubregistry(tokenId, IRegistry(transferData.newSubregistry));
+        registry.setResolver(tokenId, transferData.newResolver);
+        registry.safeTransferFrom(address(this), transferData.newOwner, tokenId, 1, "");
```
Arachnid, line 36: Thought for a future PR - should we offer a method on the registry that does all of this?

contracts/src/common/EjectionController.sol
```diff
@@ -0,0 +1,92 @@
...
+
+import {ITokenObserver} from "./ITokenObserver.sol";  
+import {IERC1155Receiver} from "@openzeppelin/contracts/token/ERC1155/IERC1155Receiver.sol";
+import {IPermissionedRegistry} from "./IPermissionedRegistry.sol";
+
+/**
+ * @title EjectionController
+ * @dev Base contract for the ejection controllers.
+ */
+abstract contract EjectionController is ITokenObserver, IERC1155Receiver {
```
Arachnid, line 12: I think the `TokenObserver` interface  can be moved to the L2 subclass.

contracts/src/L2/L2EjectionController.sol
```diff
@@ -0,0 +1,64 @@
...
+
+    /**
+     * Overrides the EjectionController._onEject function.
+     */
+    function _onEject(uint256[] memory tokenIds, TransferData[] memory transferDataArray) internal virtual override {
+        for (uint256 i = 0; i < tokenIds.length; i++) {
+            uint256 tokenId = tokenIds[i];
+            TransferData memory transferData = transferDataArray[i];
+
+            if (registry.ownerOf(tokenId) != address(this)) {
```
Arachnid, line 49: I think you can move this outside the loop, just as in the L1 version.

Arachnid (changes requested):

contracts/src/L2/L2EjectionController.sol
hiddentao: Not sure what the use case will be aside from migration. It's essentially a "transfer to someone new with these values set".

PR #40, opened 2025-04-17
Add a Dockerfile and Compose for devnet use

sugh01:


mdtanrikulu:
lgtm, just the setup file belongs to the old working commit, will require an update when ready from the original PR.

mdtanrikulu (approved):

PR #43, opened 2025-05-08
refactor: update code to work with new ejection controller logic

hiddentao:


contracts/src/mocks/MockL2Bridge.sol
```diff
@@ -48,34 +53,15 @@ contract MockL2Bridge {
         bytes4 messageType = bytes4(message[:4]);
         
         if (messageType == bytes4(keccak256("NAME_MIGRATION"))) {
```
mdtanrikulu, line 55:
```suggestion
        if (messageType == bytes4(keccak256("NAME_EJECTION"))) {
```

mdtanrikulu (changes requested):

mdtanrikulu (approved):

PR #48, opened 2025-05-15
BET-262: L1MigrationController

hiddentao:


contracts/src/common/MigrationController.sol
```diff
@@ -0,0 +1,141 @@
...
+    function setStrategy(IMigrationStrategy _strategy) external onlyOwner {
+        strategy = _strategy;
+        emit StrategySet(_strategy);
+    }
+
+    /**
+     * Implements ERC165.supportsInterface
+     */
+    function supportsInterface(bytes4 interfaceId) public virtual view override(ERC165, IERC165) returns (bool) {
+        return interfaceId == type(MigrationController).interfaceId || interfaceId == type(Ownable).interfaceId || interfaceId == type(IERC1155Receiver).interfaceId || interfaceId == type(IERC721Receiver).interfaceId || super.supportsInterface(interfaceId);
```
Arachnid, line 46: Can you format this with each condition on a line? Feel free to use a linter/prettifier instead.

contracts/src/common/MigrationController.sol
```diff
@@ -0,0 +1,141 @@
...
+     * If this is called then it means an unwrapped .eth name is being migrated to v2.
+     */
+    function onERC721Received(address /*operator*/, address /*from*/, uint256 tokenId, bytes calldata data) external virtual returns (bytes4) {
+        if (msg.sender != address(ethRegistryV1)) {
+            revert CallerNotEthRegistryV1(msg.sender);
+        }
+
+        if (ethRegistryV1.ownerOf(tokenId) != address(this)) {
+            revert NotOwner(tokenId);
+        }
```
Arachnid, line 88: If we know the caller is the registry, do we need this check?

contracts/src/common/TransferData.sol
```diff
@@ -0,0 +1,29 @@
...
+struct MigrationData {
+    TransferData transferData;
+    /**
+     * @dev If true, the name is being migrated to L1. If false, the name is being migrated to L2.
+     */
+    bool toL1;
+    /**
+     * @dev Additional data for the migration (e.g to hold proof data).
+     */
+    bytes data;
```
Arachnid, line 28: Is this needed anywhere?

Arachnid (changes requested):

contracts/src/common/TransferData.sol
hiddentao: I want to leave this in for future-proofing

Arachnid: Let's leave it out and we can add it when we write something that needs it.

hiddentao:
Latest changes:

- Refactored bridging interfaces to be more generic and better able to support migrations vs ejections. Was also able to delete redundant code as part of this
- Updated migration controller based on review feedback
- Migration controller now speaks to the bridge

contracts/src/common/IMigration.sol
```diff
@@ -0,0 +1,11 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.8.13;
+
+import {MigrationData} from "./TransferData.sol";
+
+/**
+ * @dev Interface for the migration strategies that get used by MigrationController instances.
+ */
+interface IMigrationStrategy {
+  function migrateLockedEthName(address nameWrapperRegistry, uint256 tokenId, MigrationData memory migrationData) external;
```
Arachnid, line 10: I don't think the registry address is necessary here.

contracts/src/L1/L1MigrationController.sol
```diff
@@ -0,0 +1,143 @@
...
+    function _migrateWrappedEthNames(uint256[] memory tokenIds, MigrationData[] memory migrationDataArray) internal {                
+        for (uint256 i = 0; i < tokenIds.length; i++) {
+            (, uint32 fuses, ) = nameWrapper.getData(tokenIds[i]);
+            
+            if (fuses & CANNOT_UNWRAP != 0) { // Name is locked
+                if (address(strategy) == address(0)) {
+                    revert NoMigrationStrategySet();
+                }
+                // Name is locked, migrate through strategy
+                strategy.migrateLockedEthName(address(nameWrapper), tokenIds[i], migrationDataArray[i]);
```
Arachnid, line 125: We need to think about how we'll handle migrating subdomains, if the strategy supports it, too.

contracts/src/L1/L1MigrationController.sol
```diff
@@ -0,0 +1,143 @@
...
+        }
+    }
+
+    /**
+     * @dev Migrate a name via the bridge.
+     *
+     * @param tokenId The token ID of the .eth name.
+     * @param migrationData The migration data.
+     */
+    function _migrateNameViaBridge(uint256 tokenId, MigrationData memory migrationData) internal {
```
Arachnid, line 139: We need to check that the token ID is valid for the label here.

contracts/src/mocks/MockL1Bridge.sol
```diff
@@ -1,79 +1,58 @@
...
+            revert BridgeTargetNotSupported();
+        }
+        
+        (BridgeMessageType messageType, uint256 tokenId, bytes memory data) = BridgeEncoder.decode(message);
+        
+        if (messageType == BridgeMessageType.EJECTION) {
+            emit NameEjectedToL2(tokenId, data);
+        } else if (messageType == BridgeMessageType.MIGRATION) {
+            emit NameMigratedToL2(tokenId, data);
+        }
```
Arachnid, line 40: We need an `else` case here with an error.

contracts/src/mocks/MockL1Bridge.sol
```diff
@@ -1,79 +1,58 @@
...
+        BridgeMessageType messageType,
+        uint256 /*tokenId*/,
+        bytes memory data
+    ) internal override {
+        if (messageType == BridgeMessageType.EJECTION) {
+            TransferData memory _transferData = abi.decode(data, (TransferData));
+            ejectionController.completeEjectionFromL2(_transferData);
+        } else if (messageType == BridgeMessageType.MIGRATION) {
+            // TODO: handle migration messages
         }
```
Arachnid, line 56: We need an `else` case here with a revert.

contracts/src/mocks/MockL1Bridge.sol
```diff
@@ -1,79 +1,58 @@
...
+    function _handleDecodedMessage(
+        BridgeMessageType messageType,
+        uint256 /*tokenId*/,
+        bytes memory data
+    ) internal override {
+        if (messageType == BridgeMessageType.EJECTION) {
+            TransferData memory _transferData = abi.decode(data, (TransferData));
+            ejectionController.completeEjectionFromL2(_transferData);
+        } else if (messageType == BridgeMessageType.MIGRATION) {
+            // TODO: handle migration messages
```
Arachnid, line 55: I don't think the L1 bridge can receive a migration message from L2.

contracts/src/mocks/MockL1EjectionController.sol
Arachnid, line 1: If we're defining a generic 'bridge' contract, then can we remove the need for chain-specific ejection controllers?

contracts/src/mocks/MockL2Bridge.sol
```diff
@@ -1,73 +1,61 @@
...
+        }
+        
+        (BridgeMessageType messageType, uint256 tokenId, bytes memory data) = BridgeEncoder.decode(message);
+        
+        if (messageType == BridgeMessageType.MIGRATION) {
+            // Migration messages are not supported in L2 bridge
+            revert MigrationNotSupported();
+        } else if (messageType == BridgeMessageType.EJECTION) {
+            emit NameEjectedToL1(tokenId, data);
+        }
```
Arachnid, line 43: We need an `else` with a revert here.

contracts/src/L1/L1MigrationController.sol
hiddentao: Yeah, I was going to leave this to be handled within the concerete strategy

contracts/src/common/IBridge.sol
```diff
@@ -0,0 +1,60 @@
...
+    L1,
+    L2
+}
+
+/**
+ * @dev Interface for the bridge contract.
+ */
+interface IBridge {
+    function sendMessage(BridgeTarget target, bytes memory message) external;
+    function receiveMessage(bytes memory message) external;
```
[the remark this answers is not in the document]
hiddentao: Vary as in how? Regarding the interface, the specific bridge implementation (e.g MockL1Bridge) will know how to call the specific contract (e.g MockL1EjectionController)

contracts/src/mocks/MockL1Bridge.sol
hiddentao: BridgeMessage.decode will panic in that case, so we don't need this.

contracts/src/mocks/MockL1EjectionController.sol
hiddentao: No, because the ejection controllers behave differently on each chain

contracts/src/L1/L1MigrationController.sol
```diff
@@ -0,0 +1,143 @@
+// SPDX-License-Identifier: MIT
```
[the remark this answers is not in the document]
hiddentao: we need the name to be registered on L2 even if the user wants it to be migrated to L1 - this is due to the assumptions made by our ejection as well as renewal logic. so we either spend the gas during migration or we'll have to spend the gas in the ejection controllers to check for this special case of migrated-to-L1 names. On balance I think we're better off spending that gas once during migration.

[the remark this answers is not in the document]
hiddentao: Done.

contracts/src/mocks/MockL1EjectionController.sol
```diff
@@ -1,45 +1,54 @@
...
+import {NameEjectedToL1, NameEjectedToL2} from "./MockEjectionControllerEvents.sol";
 
 /**
  * @title MockL1EjectionController
  * @dev Controller for handling L1 ENS operations with PermissionedRegistry
  */
-contract MockL1EjectionController is L1EjectionController {
-    IBridge public bridge;
+contract MockL1EjectionController is L1EjectionController, IL1Migrator {
+    MockL1Bridge public bridge;
```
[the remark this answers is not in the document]
hiddentao: Them mock ejection controllers exist because we have separate ejection controllers for L1 and L2, and that's because they actually do different things. The base ejection controller classes are abstract and only expose internal methods, leaving public/external method implementation details upto concrete classes.

We have separate mock L1 and L2 bridges because they do different things as well - their sendMessage logic is different. Of course, we could just have single mock bridge contract but then we'd need to have a bridge message target property to make the code clean - we used to have that before but you weren't keen on that since you said we're only ever going to send between L1 and L2 as targets.

contracts/src/L1/L1UnlockedMigrationController.sol
```diff
@@ -0,0 +1,147 @@
...
+contract L1UnlockedMigrationController is IERC1155Receiver, IERC721Receiver, ERC165, Ownable {
+    error UnauthorizedCaller(address caller);   
+    error MigrationFailed();
+    error TokenIdMismatch(uint256 tokenId, uint256 expectedTokenId);
+    error MigrationNotSupported();
+
+    IBaseRegistrar public immutable ethRegistryV1;
+    INameWrapper public immutable nameWrapper;
+    IBridge public immutable bridge;
+    IL1Migrator public immutable l1Migrator;
```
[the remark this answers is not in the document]
hiddentao: Well, the nice thing about having this interface and then re-using the L1 ejection controller as the implementation is that we've already got the code there to register a name on L1. If we were to pass the L1 .eth registry directly then we'd have to write that code again, and then we'd have two places doing these name registrations."

But if you still want me to make this change I can.

Arachnid (changes requested):

contracts/src/mocks/MockL1EjectionController.sol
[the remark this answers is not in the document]
hiddentao: The separate mocks are needed because the L1 and L2 bridges operate a bit differently, even the mock ones. Same for the ejection controllers. I've tried simplifying a bit already.

PR #52, opened 2025-05-19
feat: devnet + deploy scripts (hardhat-deploy)

TateB:
changes:
- devnet uses deploy scripts
- devnet uses bun
- hardhat version now on v3
- extended devnet deployment to also deploy the root registry


.github/workflows/main.yml
```diff
@@ -17,6 +17,11 @@ jobs:
       - name: Install Foundry
         uses: foundry-rs/foundry-toolchain@v1
 
+      - name: Use Node 23
+        uses: actions/setup-node@v4
+        with:
+          node-version: 23
```
talentlessguy, line 23: don't use odd versions of Node because their lifespan is really short. Switch to Node 24 or 22 instead

mdtanrikulu (approved):

adraffy:
I browsed through it and it looks pretty good.  I'm glad HH finally got away from a singular network.  I like what you did with the artifact types.

I'll have to play with a bit and get back with specific feedback.

Arachnid (changes requested):

PR #58, opened 2025-05-31
Add `DedicatedResolver`

adraffy:


contracts/src/common/DedicatedResolver.sol
```diff
@@ -0,0 +1,300 @@
...
+        emit NewDediciatedResolver(owner, _wildcard);
+        if (findResolverImpl != address(0)) {
+            _setInterface(type(IFindResolver).interfaceId, findResolverImpl);
+        }
+    }
+
+    /// @inheritdoc ERC165
+    function supportsInterface(bytes4 interfaceId) public view virtual override(ERC165) returns (bool) {
+        return interfaceId == type(IExtendedResolver).interfaceId || interfaceId == type(IDedicatedResolver).interfaceId
+            || interfaceId == type(IMulticallable).interfaceId || super.supportsInterface(interfaceId);
```
Arachnid, line 96: Please use the prettier or format these each on one line.

contracts/src/common/DedicatedResolver.sol
```diff
@@ -0,0 +1,300 @@
...
+    /// @param coinType The coin type.
+    /// @param addressBytes The address to set.
+    function setAddr(uint256 coinType, bytes memory addressBytes) external onlyOwner {
+        if (addressBytes.length != 0 && ENSIP19.isEVMCoinType(coinType)) {
+            if (addressBytes.length != 20) {
+                revert InvalidEVMAddress(addressBytes);
+            } else if (bytes20(addressBytes) == bytes20(0)) {
+                addressBytes = "";
+            }
+        }
```
Arachnid, line 122: I think we discussed this elsewhere recently - we should treat `bytes20(0)` as an explicit request to return the 0 address, while we should treat an empty string as a request to fall through to the default.

contracts/src/common/DedicatedResolver.sol
```diff
@@ -0,0 +1,300 @@
...
+        addressBytes = _addresses[coinType];
+        if (addressBytes.length == 0 && ENSIP19.chainFromCoinType(coinType) > 0) {
+            addressBytes = _addresses[EVM_BIT];
+        }
+    }
+
+    /// @notice Get `addr(60)` as `address`.
+    /// @return addr_ The address for coin type 60.
+    function addr(bytes32) public view returns (address payable addr_) {
+        addr_ = payable(address(bytes20(addr(NODE_ANY, COIN_TYPE_ETH))));
```
Arachnid, line 144: By convention, use `return` instead of named return values, unless it'd be clearer (due to control flow) to do otherwise.

contracts/src/common/DedicatedResolver.sol
```diff
@@ -0,0 +1,300 @@
...
+
+    /// @inheritdoc ERC165
+    function supportsInterface(bytes4 interfaceId) public view virtual override(ERC165) returns (bool) {
+        return interfaceId == type(IExtendedResolver).interfaceId || interfaceId == type(IDedicatedResolver).interfaceId
+            || interfaceId == type(IMulticallable).interfaceId || super.supportsInterface(interfaceId);
+    }
+
+    /// @notice Get the `IFindResolver` implementation.
+    ///         Defaults to the proxy implementation if unset.
+    function findResolverImplementation() public view returns (address impl) {
```
Arachnid, line 101: This should return `IFindResolver`

contracts/src/common/DedicatedResolver.sol
```diff
@@ -0,0 +1,300 @@
...
+    function name(bytes32) external view returns (string memory name_) {
+        name_ = _primary;
+    }
+
+    /// @inheritdoc IDedicatedResolver
+    /// @dev True if `wildcard` or this resolver is exact in the registry.
+    function supportsName(bytes memory _name) public view returns (bool) {
+        if (wildcard) return true;
+        address impl = findResolverImplementation();
+        if (impl == address(0)) return false;
```
Arachnid, line 266: I think this should throw an error? Otherwise it will just mysteriously return empty results for all records.

contracts/src/common/IFindResolver.sol
```diff
@@ -0,0 +1,6 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.8.13;
+
+interface IFindResolver {
```
Arachnid, line 4: I think something like `IResolverFinder` might be a clearer name.

Arachnid: Also, can you make the universal resolver implement this interface, so we'll get a compiler error if it breaks?

Arachnid (changes requested):

contracts/src/common/DedicatedResolver.sol
adraffy: I don't think this repo had a Solidity formatter chosen yet—this was likely `forge fmt`.  I added the same package as ens-contracts and updated the config.

contracts/src/common/DedicatedResolver.sol
adraffy: Okay, I agree with this.

contracts/src/common/DedicatedResolver.sol
adraffy: We discussed this in Slack and chose the design where a non-zero default is set by construction.  I also added a hard `error ResolverFinderRequired()` check.

contracts/src/common/IFindResolver.sol
adraffy: Changed name and added interface to AUR.

contracts/src/L1/ETHFallbackResolver.sol
```diff
@@ -152,7 +180,9 @@ contract ETHFallbackResolver is IExtendedResolver, GatewayFetchTarget, CCIPReade
     /// @dev Return true if the name is actively registered on V1.
     /// @param id The labelhash of the "eth" 2LD.
     function _isActiveRegistrationV1(uint256 id) internal view returns (bool) {
-        return !ethRegistrarV1.available(id) && ethRegistrarV1.ownerOf(id) != burnAddressV1;
+        return
+            !ethRegistrarV1.available(id) &&
+            ethRegistrarV1.ownerOf(id) != burnAddressV1;
```
Arachnid, line 185: This will do for this PR, but will likely need revising as we discuss our best options.

contracts/src/common/DedicatedResolver.sol
```diff
@@ -0,0 +1,305 @@
...
+        return results;
+    }
+
+    /// @notice Warning: node check is ignored.
+    function multicallWithNodeCheck(
+        bytes32,
+        bytes[] calldata calls
+    ) external returns (bytes[] memory) {
+        return multicall(calls);
+    }
```
Arachnid, line 304: Why?

contracts/src/common/IDedicatedResolverManager.sol
```diff
@@ -0,0 +1,47 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.8.13;
+
+/// @dev Placeholder node for compatibility with standard resolver behavior.
+bytes32 constant NODE_ANY = 0;
+
+/// @notice Interface for a resolver that returns the same results for all nodes.
+/// @dev Interface selector: `0x92349baa`
+interface IDedicatedResolverManager {
```
Arachnid, line 9: How about `IDedicatedResolverSetters`?

Arachnid (changes requested):

contracts/src/L1/ETHFallbackResolver.sol
adraffy: I could add the NameWrapper logic and tests if we want to explore that option, otherwise I'll wait.

contracts/src/common/DedicatedResolver.sol
adraffy: The node check logic asserts all nodes are the same—which ensures that a trusted writer can't clobber arbitrary records.  Since all writes to `DedicatedResolver` update the same records, this check doesn't accomplish anything.

In a prior commit, I just had `multicall()` implemented but then I couldn't satisfy the `IMulticallable` interface.

contracts/src/common/IDedicatedResolverManager.sol
adraffy: Changed

contracts/src/common/DedicatedResolver.sol
adraffy: I could add a similar explanation to the docstring.

Arachnid: Yes please!

Arachnid (approved):
Two nits; otherwise LGTM.

contracts/src/L1/ETHFallbackResolver.sol
```diff
@@ -40,30 +52,30 @@ contract ETHFallbackResolver is IExtendedResolver, GatewayFetchTarget, CCIPReade
...
-    uint256 constant SLOT_PR_PUBKEYS = 9;
-    uint256 constant SLOT_PR_TEXTS = 10;
+    /// @dev Storage layout of DedicatedResolver.
+    uint256 constant SLOT_DR_ADDRESSES = 0; // _addresses
+    uint256 constant SLOT_DR_TEXTS = 1; // _texts
+    uint256 constant SLOT_DR_CONTENTHASH = 2; // _contenthash
+    uint256 constant SLOT_DR_PUBKEY = 3; // _pubkeyX and _pubkeyY
+    uint256 constant SLOT_DR_ABIS = 5; // _abis
+    uint256 constant SLOT_DR_INTERFACES = 6; // _interfaces
+    uint256 constant SLOT_DR_NAME = 7; // _names
```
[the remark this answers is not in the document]
adraffy: Sure, I'll add.

PR #65, opened 2025-06-11
Move RegistryDatabase event to Registry

makoto:


contracts/src/common/PermissionedRegistry.sol
```diff
@@ -19,6 +19,8 @@ import {RegistryRolesMixin} from "./RegistryRolesMixin.sol";
 
 contract PermissionedRegistry is BaseRegistry, EnhancedAccessControl, IPermissionedRegistry, MetadataMixin, RegistryRolesMixin {
     event TokenRegenerated(uint256 oldTokenId, uint256 newTokenId);
+    event SubregistryUpdate(address indexed registry, uint256 indexed id, address subregistry, uint64 expiry, uint32 data);
+    event ResolverUpdate(address indexed registry, uint256 indexed id, address resolver, uint64 expiry, uint32 data);
```
Arachnid, line 23: We can remove the `registry` argument.

Arachnid (changes requested):

contracts/src/common/PermissionedRegistry.sol
makoto: PTAL

Arachnid (approved):

PR #74, opened 2025-07-04
native remappings in hardhat compile

TateB:
- bumped hardhat versions
- changed to native hardhat remappings &rarr; `ens-contracts/` artifacts are now available
- renamed `UniversalResolver` &rarr; `UniversalResolverV2` to workaround deployment conflict
- removed manually imported artifacts


adraffy (approved):
LGTM — except I can't fix the rocketh path CI error without more investigation.

PR #77, opened 2025-07-09
BET-411: handle migration of unlocked names

hiddentao:


hiddentao:
Todo: switch to using `VerifiableFactory` and `UserRegistry` in L2 migration controller

Arachnid (changes requested):

hiddentao:
updated spec:

* names will be pre-migrated to L2 and be held by bridge controller with nullified subregistry and token observer set (so, as if they're ejected)
* thus toL1 names don't even need to send to L2
* on L2 migration handler we only need to consider 2LDs

contracts/src/L2/L2BridgeController.sol
```diff
@@ -0,0 +1,281 @@
...
+    onlyBridge 
+    {
+        (uint256 tokenId,,) = registry.getNameData(transferData.label);
+
+        if (registry.ownerOf(tokenId) != address(this)) {
+            revert NotTokenOwner(tokenId);
+        }
+
+        registry.setSubregistry(tokenId, IRegistry(transferData.subregistry));
+        registry.setResolver(tokenId, transferData.resolver);
```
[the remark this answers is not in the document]
hiddentao: Hmm, we've just removed that from the migration handling based on our convo, so will assume you want this left as is. With ejections from L1 to L2 that are reversals of previous L2 to L1 ejections, the user will probably be able to point to existing subreg and resolver addresses anyway, so it's only for ejections of previously-migrated-to-L1-only names that the user would need to create new instances.

contracts/src/L1/L1UnlockedMigrationController.sol
```diff
@@ -134,14 +134,15 @@ contract L1UnlockedMigrationController is IERC1155Receiver, IERC721Receiver, ERC
...
-        // send migration data to L2
-        bytes memory dnsEncodedName = NameUtils.dnsEncodeEthLabel(migrationData.transferData.label);
-        bytes memory message = BridgeEncoder.encodeMigration(dnsEncodedName, migrationData);
-        bridge.sendMessage(message);
-
-        // if migrated to L1 then also setup the name on the L1
+        // if migrated to L1 then setup the name on the L1
         if (migrationData.toL1) {
             l1EjectionController.completeEjectionFromL2(migrationData.transferData);
+        }
```
Arachnid, line 140: Is this going to work, given the addition of `onlyBridge`?

contracts/src/L2/L2BridgeController.sol
```diff
@@ -0,0 +1,216 @@
...
+    error MigrationFailed();
+    error InvalidTLD(bytes dnsEncodedName);
+    error NameNotFound(bytes dnsEncodedName);
+    error NotTokenOwner(uint256 tokenId);
+    error TooManyRoleAssignees(uint256 tokenId, uint256 roleBitmap);
+
+    // Events
+    event MigrationCompleted(bytes dnsEncodedName, uint256 newTokenId);
+
+    bytes32 public constant ETH_TLD_HASH = keccak256(bytes("eth"));
```
Arachnid, line 36: Does this need to be public?

contracts/src/L2/L2BridgeController.sol
```diff
@@ -0,0 +1,216 @@
...
+     * 
+     * @param dnsEncodedName The DNS encoded name being migrated
+     * @param migrationData The migration data containing transfer details
+     */
+    function completeMigrationFromL1(
+        bytes memory dnsEncodedName,
+        MigrationData memory migrationData
+    ) external onlyBridge {
+        // if migrating to L1 then there is nothing to do, else let's create a subregistry
+        if (!migrationData.toL1) {
```
Arachnid, line 60: Is this field and this `if` statement still required?

contracts/src/L2/L2BridgeController.sol
```diff
@@ -0,0 +1,216 @@
...
+            // owner should be the bridge controller
+            if (registry.ownerOf(tokenId) != address(this)) {
+                revert NotTokenOwner(tokenId);
+            }
+
+            registry.setSubregistry(tokenId, IPermissionedRegistry(migrationData.transferData.subregistry));
+            registry.setResolver(tokenId, migrationData.transferData.resolver);
+
+            // now unset the token observer and transfer the name to the owner
+            registry.setTokenObserver(tokenId, ITokenObserver(address(0)));
```
Arachnid, line 73: We should probably amend the registry to clear the token observer on transfer, instead.

contracts/src/L2/L2BridgeController.sol
Arachnid: Without the `if`, it looks like the migration and ejection code might actually be identical?

contracts/src/L2/L2BridgeController.sol
```diff
@@ -0,0 +1,216 @@
...
+        // Verify TLD is "eth"
+        (bytes32 tldHash, ) = NameCoder.readLabel(name, tldOffset);
+        if (tldHash != ETH_TLD_HASH) {
+            revert InvalidTLD(name);
+        }
+        
+        // Verify the name ends after TLD (next byte should be 0)
+        if (name[tldOffset + 1 + tldSize] != 0) {
+            revert InvalidTLD(name);
+        }
```
Arachnid, line 165:
```suggestion
        // Verify the name is a .eth 2LD
        (bytes32 tldHash, ) = NameCoder.readLabel(name, tldOffset);
        if (tldHash != ETH_TLD_HASH || name[tldOffset + 1 + tldSize] != 0)) {
            revert InvalidTLD(name);
        }
```

contracts/src/L2/L2BridgeController.sol
```diff
@@ -0,0 +1,216 @@
...
+        
+        // Read the 2LD label
+        string memory label = NameUtils.readLabel(name, 0);
```
Arachnid, line 168: If you move this up to the top, you won't have to calculate the TLD offset separately.

contracts/src/L2/L2BridgeController.sol
```diff
@@ -0,0 +1,216 @@
...
+        for (uint256 i = 0; i < tokenIds.length; i++) {
+            uint256 tokenId = tokenIds[i];
+            TransferData memory transferData = transferDataArray[i];
+
+            // check that the label matches the token id
+            _assertTokenIdMatchesLabel(tokenId, transferData.label);
+
+            // check that there is no more than holder of the token observer and subregistry setting roles
+            uint256 roleBitmap = ROLE_SET_TOKEN_OBSERVER | ROLE_SET_SUBREGISTRY;
+            (uint256 counts, uint256 mask) = IEnhancedAccessControl(address(registry)).getAssigneeCount(registry.getTokenIdResource(tokenId), roleBitmap);
```
Arachnid, line 194: Perhaps we could provide an overload of `getAssigneeCount` on `PermissionedRegistry` that takes a token ID?

contracts/src/L2/L2BridgeController.sol
```diff
@@ -0,0 +1,216 @@
...
+            if (counts & mask != roleBitmap) {
```
Arachnid, line 195: A comment here on why this ensures there's exactly one holder of each of those roles would be useful. Also, we should note that we don't need to explicitly check we're the owner of those roles, since we exercise them both further down.

contracts/src/L2/L2BridgeController.sol
```diff
@@ -0,0 +1,216 @@
...
+    function _onEject(uint256[] memory tokenIds, TransferData[] memory transferDataArray) internal virtual override {
+        for (uint256 i = 0; i < tokenIds.length; i++) {
+            uint256 tokenId = tokenIds[i];
+            TransferData memory transferData = transferDataArray[i];
+
+            // check that the label matches the token id
+            _assertTokenIdMatchesLabel(tokenId, transferData.label);
+
+            // check that there is no more than holder of the token observer and subregistry setting roles
+            uint256 roleBitmap = ROLE_SET_TOKEN_OBSERVER | ROLE_SET_SUBREGISTRY;
```
Arachnid, line 193: This also needs to include the admin roles for these roles, I think.

contracts/src/common/EnhancedAccessControl.sol
```diff
@@ -120,7 +111,21 @@ abstract contract EnhancedAccessControl is Context, ERC165 {
...
+        (uint256 counts, ) = getAssigneeCount(resource, roleBitmap);
+        return counts != 0;
+    }
+
+    /**
+     * @dev Get the no. of assignees for the roles in the given role bitmap.
+     *
+     * @param resource The resource to check.
+     * @param roleBitmap The roles bitmap to check.
+     * @return counts The no. of assignees for each of the roles in the given role bitmap, expressed as a bitmap.
```
Arachnid, line 123:
```suggestion
     * @return counts The no. of assignees for each of the roles in the given role bitmap, expressed as a packed array of 4-bit ints.
```

Arachnid (changes requested):

contracts/src/L1/L1UnlockedMigrationController.sol
hiddentao: Good catch. Looks like the tests weren't doing their job properly. I've now replaced this wholesale with a role-based access control system using EAC.

contracts/src/L2/L2BridgeController.sol
hiddentao: You're right, will refactor.

contracts/src/L2/L2BridgeController.sol
```diff
@@ -0,0 +1,183 @@
...
+    function _onEject(uint256[] memory tokenIds, TransferData[] memory transferDataArray) internal virtual override {
+        for (uint256 i = 0; i < tokenIds.length; i++) {
+            uint256 tokenId = tokenIds[i];
+            TransferData memory transferData = transferDataArray[i];
+
+            // check that the label matches the token id
+            _assertTokenIdMatchesLabel(tokenId, transferData.label);
+
+            /*
+            Check that there is no more than holder of the token observer and subregistry setting roles.
```
Arachnid, line 145:
```suggestion
            Check that there is no more than one holder of the token observer and subregistry setting roles.
```

contracts/src/L2/L2BridgeController.sol
```diff
@@ -0,0 +1,183 @@
...
+            This works by calculating the no. of assignees for each of the given roles as a bitmap `(counts & mask)` where each role's corresponding 
+            nybble is set to its assignee count.
+
+            Since the roles themselves are bitmaps where each role's nybble is set to 1, we can simply comparing the two values to 
+            check to see if each role has exactly one assignee.
+
+            We also don't need to check that we (the bridge controller) are the sole assignee of these roles since we exercise these 
+            roles further down below.
+            */
+            uint256 roleBitmap =
```
Arachnid, line 156: Should we also prevent ejection if another account can set the resolver?

Arachnid (changes requested):

contracts/src/L2/L2BridgeController.sol
hiddentao: I don't see why we need that - we deliberately don't touch the resolver so that it keeps resolving however the owner wants it to.

Arachnid (approved):

PR #82, opened 2025-07-23
Feat/bet 416 l2 registrar erc20

makoto:
## Changes

  - Removed mutable functions: setPriceOracle and setCommitmentAges to make ETHRegistrar immutable
  - Added SafeERC20 usage for secure token transfers with proper error handling
  - Beneficiary payment system - all fees now flow directly to designated `beneficiary` address as users register/renew. No `withdraw` function
  - USD-denominated pricing with automatic token conversion (6-decimal precision)
  - Token-agnostic events - emit USD prices regardless of payment token
  - Dynamic pricing support through configurable rent price arrays

  Technical Achievements

##  Security Considerations

  - `ETHRegistrar` is now immutable. To upgrade, deploy new contract
  - We assume the tokens and beneficiary addresses are trusted hence no special security consideration. However there is likelyhood that the sender or beneficiary addresses could get blacklisted for stables in future. The workaround is to always accept multiple token addresses to avoid single point of failure.

## Out of scope

- dynamic pricing
- premium pricing

## Open for discussion

- Currently USD decimal is set to 6 (same as USDC). Should we change to 2?


hiddentao (changes requested):
Overall good, just a few minor changes to be made.

contracts/src/L2/ETHRegistrar.sol
hiddentao, line 1: The constructor should take an `ITokenPriceOracle` as the argument type instead of `IPriceOracle` - this then removes the need for casting and keeps registrar bytecode minimal.

hiddentao (changes requested):
ETHRegistrar can be simplified further, see comment

contracts/src/L2/ETHRegistrar.sol
```diff
@@ -151,60 +171,67 @@ contract ETHRegistrar is IETHRegistrar, EnhancedAccessControl {
...
-
-        emit NameRenewed(name, duration, tokenId, newExpiry);
+        uint64 newExpiry = expiry + duration;
+        
+        // Get USD pricing breakdown
+        ITokenPriceOracle.Price memory usdPrice = prices.price(name, uint256(expiry), duration);
+        
+        // Convert to token amount for payment and handle transfer
+        {
+            ITokenPriceOracle tokenOracle = prices;
```
hiddentao, line 218: Redundant. Can use it directly - perhaps we should just call the member variable to tokenOracle instead of prices. 

Actually let's call it `tokenPriceOracle` instead of prices.

contracts/src/L2/ETHRegistrar.sol
```diff
@@ -5,26 +5,22 @@ import {IETHRegistrar} from "./IETHRegistrar.sol";
 import {IRegistry} from "../common/IRegistry.sol";
 import {IERC1155Singleton} from "../common/IERC1155Singleton.sol";
 import {IPermissionedRegistry} from "../common/IPermissionedRegistry.sol";
-import {IPriceOracle} from "@ens/contracts/ethregistrar/IPriceOracle.sol";
+import {ITokenPriceOracle} from "./ITokenPriceOracle.sol";
+import {TokenPriceOracle} from "./TokenPriceOracle.sol";
```
hiddentao, line 9: Why is this import needed?

hiddentao (approved):

contracts/src/L2/TokenPriceOracle.sol
```diff
@@ -0,0 +1,155 @@
...
+        view 
+        virtual 
+        returns (uint256) 
+    {
+        // Default implementation: no premium
+        return 0; // $0 in USD_DECIMALS
+    }
+
+
+    function supportsInterface(bytes4 interfaceId) public view virtual returns (bool) {
```
adraffy, line 150: Oh oops, I had a pending review but didn't finish it.

Nick suggested to me that we should use ERC165 instead of a direct implementation.

PR #87, opened 2025-07-29
BET-429: Exclude mocks and test code from coverage reports

hiddentao:


makoto:
Manually verified that it's now excluding mock and test directory
Screenshot 2025-07-29 at 18 04 47

PR #89, opened 2025-07-30
BET-430: max role assignee counts no longer prevent name transfers

hiddentao:


contracts/src/common/EnhancedAccessControl.sol
```diff
@@ -226,16 +226,24 @@ abstract contract EnhancedAccessControl is Context, ERC165, IEnhancedAccessContr
...
+     * @param resource The resource to transfer roles within.
+     * @param srcAccount The account to transfer roles from.
+     * @param dstAccount The account to transfer roles to.
      * @param executeCallbacks Whether to execute the callbacks.
      */
-    function _copyRoles(uint256 resource, address srcAccount, address dstAccount, bool executeCallbacks) internal virtual {
+    function _transferRoles(uint256 resource, address srcAccount, address dstAccount, bool executeCallbacks) internal virtual {
         uint256 srcRoles = _roles[resource][srcAccount];
-        _grantRoles(resource, srcRoles, dstAccount, executeCallbacks);
+        if (srcRoles != 0) {
```
makoto, line 241: Should we have a guard against transfering to 0x address? If people keep transferring roles to 0x, no one can revoke from 0x hence eventually hitting the maxroles?

hiddentao: Transferring to 0x is actually part of the token regeneration process since we burn then mint.

PR #103, opened 2025-08-14
fix: price oracle deployment

TateB:
price oracle was previously being deployed with $5/second pricing for 3/4/5 char names. changed to pre-existing pricing in ENSv1 + added nd/s util funcs


f1lander (commented):
So I tested this locally, is returning like 4 thousand  formatted the price by a year with 6 units, changed to 9 decimals in the TokenPriceOracle contract is returning the correct amount of 4.98 usdc 

@TateB

PR #105, opened 2025-08-16
Cleanup `ETHRegistrar`

adraffy:
- removed `TokenPriceOracle`, `StablePriceOracle` and related tests

```
ETHRegistrar ==> IRentPriceOracle
                       \/
             StandardRentPriceOracle 
```
- added [IRentPriceOracle.sol](https://github.com/ensdomains/namechain/blob/fix/redesign-eth-registrar/contracts/src/L2/IRentPriceOracle.sol)
- added [StandardRentPriceOracle.sol](https://github.com/ensdomains/namechain/blob/fix/redesign-eth-registrar/contracts/src/L2/StandardRentPriceOracle.sol) and [tests](https://github.com/ensdomains/namechain/blob/fix/redesign-eth-registrar/contracts/test/StandardRentPriceOracle.t.sol)
    * takes `PaymentRatio(ERC20, numer, denom)[]` in constructor
    * changed `basePrice()` to `baseRate()`
    * changed `isValid()` to `baseRate() > 0` instead of `strlen() >= 3`
   * ~~added grace period~~ &rarr; no premium for latest owner
    * added premium w/exponential decay
    * added `updateBaseRates()`
    * added `updatePremumPricing()`
    * added `updatePaymentToken()`
- changed [IETHRegistrar.sol](https://github.com/ensdomains/namechain/blob/fix/redesign-eth-registrar/contracts/src/L2/IETHRegistrar.sol)
    * moved events from `ETHRegistrar`
    * added `payableToken` and `referer` to events, and reordered
    * changed `base` and `premium` to use token amounts
    * added `commitmentAt()` so it's part of the interface
    * extends `IRentPriceOracle` 
- redesigned [ETHRegistrar.sol](https://github.com/ensdomains/namechain/blob/fix/redesign-eth-registrar/contracts/src/L2/ETHRegistrar.sol)
   * `setRentPriceOracle()` so pricing logic can be upgraded
   * `{isValid, isPaymentToken, rentPrice}` forward to `IRentPriceOracle`
   * made almost everything configurable via constructor or `IRentPriceOracle`
   * added `referer`
- redesigned [ETHRegistrar.t.sol](https://github.com/ensdomains/namechain/blob/fix/redesign-eth-registrar/contracts/test/ETHRegistrar.t.sol)
   * significantly simpler
   * merged in `ETHRegistrarOperationalRisks.t.sol`
   * removed `SafeERC20` and tests (Tate) but kept `require()` on return value (Nick)
- added `IPermissionedRegistry.latestOwnerOf()` and some tests
    * does NOT survive an explicit `burn()`
- added [HalvingUtils.sol](https://github.com/ensdomains/namechain/blob/fix/redesign-eth-registrar/contracts/src/common/HalvingUtils.sol) and [tests](https://github.com/ensdomains/namechain/blob/fix/redesign-eth-registrar/contracts/test/HalvingUtils.t.sol)
- added [StandardPricing.sol](https://github.com/ensdomains/namechain/blob/fix/redesign-eth-registrar/contracts/test/StandardPricing.sol)

---

- ~~fixed `rmdir` &rarr; `rm`~~
- merged https://github.com/ensdomains/namechain/pull/107 &rarr; [testNames.ts](https://github.com/ensdomains/namechain/blob/fix/redesign-eth-registrar/contracts/script/testNames.ts)
- ~~fixed "excessively deep" typescript issue~~
- ~~temporarily removed [foundry.lock](https://github.com/ensdomains/namechain/blob/fix/redesign-eth-registrar/contracts/.gitignore#L31-L35)~~
- ~~added [deployVerifiableProxy.ts](https://github.com/ensdomains/namechain/blob/fix/redesign-eth-registrar/contracts/test/fixtures/deployVerifiableProxy.ts) — shared between rocketh and hardhat~~
- removed `script/readConfig.ts` — not used
- updated readme
- ~~bumped `rocketh` to 0.13+~~
- ~~removed `patches/`~~

---

* ~~do we want `minRenewDuration`? — using `1 sec`~~
* unsure  on `StandardRentPriceOracle` name? `RentPriceOracle`, `ExponentialRentPriceOracle`, `HalvingPremiumRentPriceOracle` — is oracle even the right word?
* unsure on `RentPriceOracleChanged` event name and args — currently empty
* ~~possibly `paymentTokens` should be iterable?~~


contracts/src/L2/ETHRegistrar.sol
```diff
@@ -1,271 +1,356 @@
...
         }
+        _;
+    }
 
-        tokenPriceOracle = _prices;
-        minCommitmentAge = _minCommitmentAge;
-        maxCommitmentAge = _maxCommitmentAge;
-        beneficiary = _beneficiary;
+    /// @inheritdoc IETHRegistrar
+    function isPaymentToken(
```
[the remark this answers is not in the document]
adraffy: So it can be part of the interface, same with `committedAt()`.  For some reason, public properties don't satisfy the interface according to the compiler.

contracts/src/L2/ETHRegistrar.sol
```diff
@@ -1,271 +1,367 @@
...
+        uint64 minCommitmentAge;
+        uint64 maxCommitmentAge;
+        uint64 minRegistrationDuration;
+        uint8 priceDecimals;
+        uint256[5] baseRatePerCp;
+        uint64 premiumPeriod;
+        uint64 premiumHalvingPeriod;
+        uint256 premiumPriceInitial;
+        ITokenPriceOracle tokenPriceOracle;
+        IERC20Metadata[] paymentTokens;
```
Arachnid, line 37: I feel like this is stuff that should be abstracted out into an upgradeable contract that handles payment and pricing.

contracts/src/L2/ETHRegistrar.sol
```diff
@@ -1,271 +1,367 @@
...
+    ) public view override(EnhancedAccessControl) returns (bool) {
+        return
+            interfaceId == type(IETHRegistrar).interfaceId ||
+            super.supportsInterface(interfaceId);
+    }
 
-        registry = IPermissionedRegistry(_registry);
+    // function setPaymentToken(IERC20Metadata paymentToken, bool supported) external onlyRootRoles(LibRegistryRoles.ROLE_CONFIGURE_ADMIN) {
+    // 	_setPaymentToken(token, supported);
+    // }
```
Arachnid, line 96: Why is this here, and commented out?

contracts/src/L2/ETHRegistrar.sol
```diff
@@ -1,271 +1,367 @@
...
-     * @param name The name to check the price for.
-     * @param duration The duration of the registration or renewal.
-     * @param token The ERC20 token address.
-     * @return tokenAmount The amount of tokens required.
-     */
-    function checkPrice(string memory name, uint256 duration, address token) public view returns (uint256 tokenAmount) {
-        if (!tokenPriceOracle.getTokenConfig(token).enabled) {
-            revert TokenNotSupported(token);
+    /// @notice Get base price to register or renew `label` for `duration`.
+    ///         Use `duration = 1` for rate (price/sec).
```
Arachnid, line 143: Why not just omit the duration and let the caller multiply?

contracts/src/L2/ETHRegistrar.sol
adraffy: I didn't know if we wanted this thing to be mutable or not.

contracts/src/L2/ETHRegistrar.sol
adraffy: Here is an attempt at splitting it into two: https://github.com/ensdomains/namechain/pull/111

contracts/src/L2/ETHRegistrar.sol
adraffy: Removed.  Now immutable.  Changable via `setRentPriceOracle()`.

adraffy:
I think this isn't right, and the logic should be configurable in the `IRentPriceOracle` but what logic do I pass?

https://github.com/ensdomains/namechain/blob/97bd0086ceac269464c8b717d509011d995b9059/contracts/src/L2/ETHRegistrar.sol#L96-L103

* `expiry` (current)
* `owner` and `expiry` (new idea)
* `registry`, `token`, `expiry`?
* `isSameOwner`, expiry`
* `owner` works but then the `IRentPriceOracle` also needs to know of `ETHRegistry` — maybe that's okay?

contracts/src/L2/ETHRegistrar.sol
```diff
@@ -1,271 +1,257 @@
...
+        if (_isAvailable(oldExpiry)) {
+            revert NameNotRegistered(label);
         }
-
-        // Commit must not be too old
-        uint256 thisCommitmentValidTo = commitments[commitment] + maxCommitmentAge;
-        if (thisCommitmentValidTo <= block.timestamp) {
-            revert CommitmentTooOld(commitment, thisCommitmentValidTo, block.timestamp);
+        if (duration == 0) {
+            revert DurationTooShort(duration, 1); // minRenewDuration?
```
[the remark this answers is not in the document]
adraffy: Sure, removed.

contracts/src/L2/StandardRentPriceOracle.sol
```diff
@@ -0,0 +1,136 @@
...
+        IERC20Metadata paymentToken
+    ) public view returns (uint256 base, uint256 premium) {
+        if (!_isPaymentToken[paymentToken]) {
+            revert PaymentTokenNotSupported(paymentToken);
+        }
+        base = baseRate(label);
+        if (base == 0) {
+            revert NotRentable(label);
+        }
+        if (owner != address(0)) {
```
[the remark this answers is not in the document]
adraffy: No, since you can speculatively call this function to get the `rentPrice()` for a name when the owner isn't known, eg. what is the full price?

However, `latestOwnerOf()` can be null, so you might incorrectly get the premium discount.

contracts/src/L2/StableTokenPriceOracle.sol
```diff
@@ -0,0 +1,23 @@
...
+import {IERC20Metadata} from "@openzeppelin/contracts/token/ERC20/extensions/IERC20Metadata.sol";
+import {Math} from "@openzeppelin/contracts/utils/math/Math.sol";
+
+import {ITokenPriceOracle} from "./ITokenPriceOracle.sol";
+
+/// @notice Oracle that assumes every token is a stable coin.
+contract StableTokenPriceOracle is ITokenPriceOracle {
+    function getTokenAmount(
+        uint256 price,
+        uint8 decimals,
```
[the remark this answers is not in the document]
Arachnid: Basically, we should be treating everything as base units, and not caring about decimals anywhere. Prices should be calculated in a fixed base unit we use internally, and then converted to the amount for a specific token by fetching its exchange rate (which can be constant) as a rational number.

contracts/src/L2/ETHRegistrar.sol
```diff
@@ -1,271 +1,257 @@
...
+                duration
+            )
+        );
+        (uint256 base, uint256 premium) = rentPrice(
+            label,
+            owner,
+            duration,
+            paymentToken
+        ); // reverts if !isValid or !isPaymentToken
+        paymentToken.transferFrom(_msgSender(), beneficiary, base + premium); // reverts if payment failed
```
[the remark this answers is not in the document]
adraffy: Added`require()` since likely we'll get a typed revert from the ERC-20.

contracts/src/L2/StableTokenPriceOracle.sol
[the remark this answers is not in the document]
adraffy: `StandardRentPriceOracle` has `baseRate()` and `premiumPrice()` (not part of standard interface) which return the base units and `priceDecimals()`. 

To convert those to formatted base price. eg. USD: `premiumPrice() / 10^priceDecimals()`

It seems reasonable to change `priceDecimals()` to `priceDenominator()` or something, but I feel like this value should be exposed rather than hidden.

So if I change `uint8 decimals` to `uint256 priceDenominator`, does that address the issue?

`StableTokenPriceOracle` is certainly not the final oracle design, it's just a placeholder.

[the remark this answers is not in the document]
adraffy: Okay, I removed `ITokenPriceOracle` and `StableTokenPriceOracle`.  I removed `priceDecimals()` too.

`StandardRentPriceOracle` constructor now takes `PaymentRatio(token, numer, denom)`

contracts/src/L2/StandardRentPriceOracle.sol
```diff
@@ -0,0 +1,136 @@
...
+            // prior owner pays no premium
+            (uint256 tokenId, uint64 expiry, ) = registry.getNameData(label);
+            if (owner != registry.latestOwnerOf(tokenId)) {
+                premium = premiumPrice(expiry);
+            }
+        }
+        uint256 total = base * duration + premium; // revert on overflow
+        uint256 amount = Math.mulDiv(total, ratio.numer, ratio.denom);
+        premium = Math.mulDiv(amount, premium, total);
```
Arachnid, line 101: We shouldn't be using the same variable to represent two different quantities (and units) in different places. Why not compute the final premium directly on line 96?

contracts/src/L2/StandardRentPriceOracle.sol
```diff
@@ -0,0 +1,136 @@
...
+        base = amount - premium; // ensure: f(a+b) - f(a) == f(b)
+    }
+
+    /// @notice Get base rate to register or renew `label` for 1 second.
+    /// @param label The name to price.
+    /// @return The base rate or 0 if not rentable, in base units.
+    function baseRate(string memory label) public view returns (uint256) {
+        uint256 ncp = StringUtils.strlen(label);
+        if (ncp == 0) return 0;
```
Arachnid, line 110: Will this allow someone to register the unique zero-length name for free?

contracts/src/L2/ETHRegistrar.sol
```diff
@@ -1,271 +1,241 @@
...
-contract ETHRegistrar is IETHRegistrar, EnhancedAccessControl {
-    using SafeERC20 for IERC20;
-    uint256 private constant REGISTRATION_ROLE_BITMAP = 
-        LibRegistryRoles.ROLE_SET_SUBREGISTRY | 
-        LibRegistryRoles.ROLE_SET_SUBREGISTRY_ADMIN | 
-        LibRegistryRoles.ROLE_SET_RESOLVER | 
-        LibRegistryRoles.ROLE_SET_RESOLVER_ADMIN;
-
-    uint256 public constant MIN_REGISTRATION_DURATION = 28 days;
+uint256 constant REGISTRATION_ROLE_BITMAP = LibRegistryRoles
```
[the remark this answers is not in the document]
adraffy: Personally, I think the formatter is ugly but that's what it likes.

contracts/src/L2/StandardRentPriceOracle.sol
adraffy: Simplified the conversion logic, since it rounds up, I just did 2 ratio conversions.

contracts/src/L2/ETHRegistrar.sol
```diff
@@ -1,271 +1,241 @@
...
-        registry.renew(tokenId, newExpiry);
-        
-        emit NameRenewed(name, duration, tokenId, newExpiry, usdPrice.base);
-    }
-
-
-    function supportsInterface(bytes4 interfaceID) public view override(EnhancedAccessControl) returns (bool) {
-        return interfaceID == type(IETHRegistrar).interfaceId || super.supportsInterface(interfaceID);
+        _consumeCommitment(
+            makeCommitment(
```
TateB, line 171: add referrer to commitment

TateB (approved):

contracts/src/L2/IETHRegistrar.sol
```diff
@@ -1,133 +1,162 @@
...
+    /// @param premium The premium price, relative to `paymentToken`.
+    event NameRegistered(
+        uint256 indexed tokenId,
+        string label,
+        address owner,
+        IRegistry subregistry,
+        address resolver,
+        uint64 duration,
+        IERC20 paymentToken,
+        bytes32 referer,
```
TateB, line 71: nit: referrer misspelt (also elsewhere)

TateB (approved):

PR #113, opened 2025-08-26
Ft/debug utils

f1lander:
image

### Debug feats

This intro some debug features.


f1lander:
@mdtanrikulu something is happenning but I couldn'r figure it out with the deploykit aa script. 
image

alto/alto-config-l2-local.json
f1lander, line 1: I put this port because this changes on devnet.

PR #119, opened 2025-09-15
Fix devnet ports

adraffy:
- fix devnet ports
- bump `bun`
- added `.dockerignore`
- removed solc `0.4.11`


adraffy:
* ports and solc fixed in https://github.com/ensdomains/namechain/pull/105
* `.dockerignore` can be revisited in future for local builds
* `bun` bump was related to a docker outage

PR #128, opened 2025-09-24
feat: lint+format config

TateB:


adraffy (dismissed):
I'm still not a fan of any Solidity formatters but this seems like an improvement.

Possibly we should delete `/// @title ...` as it seems like noise when it can be derived from the first code unit or file name.

PR #147, opened 2025-10-07
Add architecture and contract page

makoto:


contracts/README.md
```diff
@@ -3,15 +3,396 @@
...
+- **Flexible Ownership**: Custom registry implementations for different ownership models
+- **Cross-Chain Support**: Seamless name resolution across L1 and L2
+- **Backward Compatibility**: Unmigrated ENSv1 names continue to function
+- **Enhanced Security**: Names can be "ejected" to L1 for maximum security guarantees
+
+### Key Features
+
+1. **Hierarchical Registries**: Each name has its own registry contract managing its subdomains
+2. **Canonical ID System**: Stable storage identifiers enabling token regeneration
+3. **Bitmap-Based Access Control**: Gas-efficient permission management supporting up to 32 roles
```
hiddentao, line 39: Replace "permission management" with "access control"

contracts/README.md
```diff
@@ -3,15 +3,396 @@
...
+- **L2 Scalability**: .eth names registered and managed on Layer 2 with reduced gas costs
+- **Flexible Ownership**: Custom registry implementations for different ownership models
+- **Cross-Chain Support**: Seamless name resolution across L1 and L2
+- **Backward Compatibility**: Unmigrated ENSv1 names continue to function
+- **Enhanced Security**: Names can be "ejected" to L1 for maximum security guarantees
+
+### Key Features
+
+1. **Hierarchical Registries**: Each name has its own registry contract managing its subdomains
+2. **Canonical ID System**: Stable storage identifiers enabling token regeneration
```
hiddentao, line 38: Doesn't quite explain what it's for. Canonical internal token ID enables for external token ID to be changed but still map to the same internal data.

contracts/README.md
```diff
@@ -3,15 +3,396 @@
...
+
+**Resolution Process**
+1. Start at root registry
+2. Recursively traverse to find the deepest registry with a resolver set
+3. Query that resolver for the requested record
+4. Supports wildcard resolution (parent resolver handles subdomains)
+
+### Token ID System
+
+ENSv2 uses a dual-identifier system to enable token regeneration while maintaining stable storage:
```
hiddentao, line 63: "dual-identifier" is confusing. Use "internal canonical ID system"

contracts/README.md
```diff
@@ -3,15 +3,396 @@
...
+
+#### How Token IDs Work
+
+```solidity
+// Label to Token ID conversion
```
hiddentao, line 68: Remove this block, users don't need to know the format of token ids - they should use `LibLabel` instead.

contracts/README.md
```diff
@@ -3,15 +3,396 @@
...
+bytes32 labelHash = keccak256(bytes(label));
+uint256 tokenId = uint256(labelHash);
+
+// Canonical ID derivation (XOR with lower 32 bits)
+uint256 canonicalId = tokenId ^ uint32(tokenId);
+```
+
+#### Key Properties
+
+1. **Token ID** (`uint256`): Unique ERC1155 identifier that can change over time
```
hiddentao, line 79: Not quite right. The system we have ensures that if we need to regenerate a token ID for a given name that the new token ID still internally maps to the data for the name. It's not that we're trying to have multiple token IDs for a given name. 

There is only one case in which we regenerate the token ID for a name - and that's when the access control settings for a name are changed. This is to prevent griefing attacks wherein a name is put on sale and - whilst it's in the process of being sold - the owner updates the access control settings on the name. To prevent this we have implemented a mechanism whereby the name's token ID gets regenerated whenever its access control settings are changed.

contracts/README.md
```diff
@@ -3,15 +3,396 @@
...
+
+// After expiry and re-registration (different tokenId!)
+uint256 tokenId2 = 0x9c22ff5f21f0b81b113e63f7db6da94fedef11b212345678901234567890abcd;
+uint256 canonical = 0x9c22ff5f21f0b81b113e63f7db6da94fedef11b200000000000000000000000; // SAME!
+
+// All storage lookups use canonical ID
+datastore.getEntry(registry, canonical); // Always finds the current data
+```
+
+**Enables**:
```
hiddentao, line 104: Remove this block

contracts/README.md
```diff
@@ -3,15 +3,396 @@
...
+// Get canonical version of any ID
+uint256 canonical = LibLabel.getCanonicalId(someTokenId);
+
+// Storage always uses canonical IDs
+DATASTORE.setSubregistry(canonicalId, subregistryAddr);
+```
+
+### Access Control
+
+ENSv2 uses **Enhanced Access Control (EAC)**, a bitmap-based permission system that stores up to 32 roles and their assignee counts in a single uint256:
```
hiddentao, line 127: replace "permission" with "role-based access control".

Upto 32 roles are allowed, with each role have a complementary _admin_ role that grants the holder permission to grant/revoke that role. Roles are assigned against a "resource", which is just a `uint256` that can represent anything. For a given role assigned for a given resource, a maximum of 15 assignees are allowed. For registered names the canonical token ID acts as the equivalent access control resource id.

contracts/README.md
```diff
@@ -3,15 +3,396 @@
...
+From [`RegistryRolesLib.sol`](src/common/registry/libraries/RegistryRolesLib.sol):
+
+| Role | Bit Position | Description |
+|------|--------------|-------------|
+| `ROLE_OWNER` | 0 | Full control over the name |
+| `ROLE_SET_RESOLVER` | 1 | Can change the resolver address |
+| `ROLE_SET_RESOLVER_ADMIN` | 2 | Can grant/revoke ROLE_SET_RESOLVER |
+| `ROLE_SET_SUBREGISTRY` | 3 | Can change subregistry addresses |
+| `ROLE_SET_SUBREGISTRY_ADMIN` | 4 | Can grant/revoke ROLE_SET_SUBREGISTRY |
+| `ROLE_CAN_TRANSFER` | 5 | Can transfer name ownership |
```
hiddentao, line 182: there is no `ROLE_CAN_TRANSFER` role.

contracts/README.md
```diff
@@ -3,15 +3,396 @@
...
+Universal storage contract for all registries. Reduces storage proof complexity for L2 resolution.
+
+**Key Functions**:
+- `getEntry(address registry, uint256 tokenId)`: Fetch subregistry and resolver
+- `setSubregistry(uint256 tokenId, address subregistry)`: Update subregistry (caller must be registry)
+- `setResolver(uint256 tokenId, address resolver)`: Update resolver
+
+**Storage Structure**:
+```solidity
+struct Entry {
```
hiddentao, line 251: This structure is incorrect

contracts/README.md
```diff
@@ -3,15 +3,396 @@
...
+}
+```
+
+#### `PermissionedRegistry` - Standard Implementation
+[src/common/registry/PermissionedRegistry.sol](src/common/registry/PermissionedRegistry.sol)
+
+Feature-complete registry with role-based access control:
+- ERC1155 NFT for subdomains
+- Enhanced Access Control with 32 roles
+- Token regeneration support
```
hiddentao, line 264: Remove this.

contracts/README.md
```diff
@@ -3,15 +3,396 @@
...
+- Fiat-pegged via Chainlink oracles
+- Multi-token support
+
+#### `L2EjectionController` - Name Migration
+[src/L2/bridge/L2EjectionController.sol](src/L2/bridge/L2EjectionController.sol)
+
+Manages ejection of names from L2 to L1:
+```solidity
+// Eject name to L1 for enhanced security
+ejectionController.initiateEjection(
```
hiddentao, line 320: There is no `initiateEjection`. Ejection controllers are now known as bridge controllers and - for both L1 and L2 - operate by having name NFTs transferred to them.

contracts/README.md
```diff
@@ -3,15 +3,396 @@
...
+
+#### `BridgeEncoder` - Message Format
+[src/common/bridge/BridgeEncoder.sol](src/common/bridge/BridgeEncoder.sol)
+
+Encodes/decodes cross-chain messages:
+
+**Message Types**:
+1. **EJECTION**: Transfer name from L2 to L1
+   ```solidity
+   bytes memory msg = BridgeEncoder.encodeEjection(
```
hiddentao, line 365: dnsEncodedName is no longer a param

contracts/README.md
makoto: I changed to this

### Mutable Token ID System

Token IDs regenerate on significant state changes (expiry, permission updates) to prevent griefing attacks from expired token approvals. There is an internal id called canonical ID that points to the same storage.

contracts/README.md
makoto: Added `Token ID Changes` and `Which Functions Use Which ID?` section

PR #253, opened 2026-03-19
Refactor Registry Metadata

adraffy:
- removed `IRegistryMetadata`
- removed `MetadataMixin`
- removed `BaseUriRegistryMetadata`
- removed `SimpleRegistryMetadata`
- removed metadata from all constructors
- added `IRegistryURIRenderer`
- changed `PermissionedRegistry`
    * added `ROLE_SET_URI` and `ROLE_SET_URI_ADMIN`
    * added `setURI(uri, renderer)`
    * added `event URIChanged(uri, renderer, sender)`
    * added `uri(tokenId)` to use renderer if token not expired


adraffy:
Not sure what [this logic](https://github.com/ensdomains/contracts-v2/blob/1c338dc46f9715ce0b947bc1928ad2e3b77f5e04/contracts/src/registry/PermissionedRegistry.sol#L273-L281) should actually be.

`return (renderer and token exists) ? renderer() : uri`

---

Edit: discussed and simplified to just: `return renderer ? renderer() : uri`

TateB (approved):

PR #259, opened 2026-03-31
feat(premigration): improve security, performance, and reliability

hiddentao:
- Support private key via PREMIGRATION_PRIVATE_KEY env var to avoid exposing secrets in process listings and shell history
- Replace sequential RPC calls with viem multicall batching, reducing 2*N round-trips per batch to just 2
- Add binary-search batch fallback to isolate failing names in ~log2(N) transactions instead of N individual ones
- Add pre-flight gas estimation with dynamic batch splitting when estimated gas exceeds 80% of block gas limit
- Add comprehensive e2e integration tests covering env var auth, multicall verification, batch sizing, checkpoint tracking, and edge cases (34 tests, all passing)
- Fix pre-existing TS errors in migration.test.ts (abitype import, hex string types)

### DIFF
--- a/contracts/script/preMigration.ts
+++ b/contracts/script/preMigration.ts
@@ -1089,7 +1089,9 @@ export async function main(argv = process.argv): Promise<void> {
     limit: opts.limit ? parseInt(opts.limit) : null,
     dryRun: opts.dryRun,
     continue: opts.continue,
-    minExpiryDays: parseInt(opts.minExpiryDays) || 7,
+    minExpiryDays: Number.isNaN(parseInt(opts.minExpiryDays))
+      ? 7
+      : parseInt(opts.minExpiryDays),
     v1ResolverAddress: opts.v1Resolver as Address,
     v1BaseRegistrarAddress: opts.v1BaseRegistrar as Address,
   };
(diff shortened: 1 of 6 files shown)


contracts/script/preMigration.ts
```diff
@@ -482,8 +542,12 @@ async function fetchAndReserveInBatches(
       let invalidLabelsInBatch = 0;
       let lastInvalidLineNumber = checkpoint.lastProcessedLineNumber;
       const validBatch = batch.filter((reg) => {
-        if (!reg.labelName || typeof reg.labelName !== 'string' || reg.labelName.trim() === '') {
-          logger.skippingInvalidName(reg.labelName || 'unknown');
+        if (
+          !reg.labelName ||
+          typeof reg.labelName !== "string" ||
+          reg.labelName.trim() === ""
+        ) {
```
adraffy: Should this check for `Buffer.from(reg.labelName).length > 255` ?

And should this check for `"[\hex{64}]"` (encoded labels) ?

TateB:
@codex

adraffy (approved):
Should we add an option to apply the [small padding](https://www.notion.so/enslabs/ENSv2-Migration-Case-Study-31f7a8b1f0ed802db30bea0050ec3b89?source=copy_link#31f7a8b1f0ed804eabc5d603a358ec13) to nearly expired names?
> Expiries “too close” (TBD: 7 days?) to ENSv2 launch will be extended in ENSv1

PR #272, opened 2026-04-14
Close role-revoke front run transfer

stevieraykatz:
When an admin broadcasts a `revokeRoles()` transaction to remove a user's roles (e.g., `ROLE_CAN_TRANSFER_ADMIN`), the user can front-run by transferring the token to another address they control. The transfer atomically moves ALL roles to the recipient via `_transferRoles()`, causing the admin's revocation to become a no-op (returns false, no revert). The recipient retains all original roles, completely bypassing admin governance.

### DIFF
--- a/contracts/src/registry/PermissionedRegistry.sol
+++ b/contracts/src/registry/PermissionedRegistry.sol
@@ -88,6 +88,8 @@ contract PermissionedRegistry is
     string internal _childLabel;
     /// @dev The entries of this registry.
     mapping(uint256 storageId => Entry entry) internal _entries;
+    /// @dev Per-name transfer lock. When set, ERC1155 transfers are blocked regardless of per-account roles.
+    mapping(uint256 storageId => bool locked) private _transferLocks;
 
     ////////////////////////////////////////////////////////////////////////
     // Initialization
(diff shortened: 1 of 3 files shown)


adraffy:
Anytime ROOT_RESOURCE has overlapping roles with a token, the registry itself is not-emancipated "in some sense".

The registry for "eth" will have the following properties:
* technically emancipated &mdash; `roles(ROOT, all accounts) & roles(any token, all accounts) == 0`
* standard emancipated &mdash; token owner is the only account with admin roles for the token-specific actions (SET_RESOLVER, SET_SUBREGISTRY, CAN_TRANSFER).

---

There's no convenience getter for this (there was one in an [older PR](https://github.com/ensdomains/contracts-v2/pull/164/changes)) but it can be computed from:

```Solidity
(uint256 rootCounts, ) = getAssigneeCount(ROOT, ALL_ROLES)
(uint256 tokenCounts, ) = getAssigneeCount(token, ALL_ROLES)

rootCount = smush(fromCount(rootCount)) // role bit set if anyone
tokenCount = smush(fromCount(tokenCounts)) // has that role

isTokenEmancipated = rootCount & tokenCount == 0; // no overlap in roles

/// @notice Merge admin and normal roles into normal roles.
function smush(uint256 roleBitmap) internal pure returns (uint128) {
     return uint128((roleBitmap | (roleBitmap << 128)) >> 128);
}

/// @notice Derive roles bitmap union for all accounts in a resource.
function fromCount(uint256 count) internal pure returns (uint256) {
    return
        (count & ALL_ROLES) |
        ((count >> 1) & ALL_ROLES) |
        ((count >> 2) & ALL_ROLES) |
        ((count >> 3) & ALL_ROLES);
}
```

makoto:
Thank you for reporting this. After discussing within the team we decided to close this as non issue

PR #274, opened 2026-04-16
Fix `ResolverProfileRewriterLib` memory overflow

adraffy:
- fixed `ResolverProfileRewriterLib.replaceNode()`
     * avoids writing if out of bounds
- added tests for malicious calldata

### DIFF
--- a/contracts/src/resolver/libraries/ResolverProfileRewriterLib.sol
+++ b/contracts/src/resolver/libraries/ResolverProfileRewriterLib.sol
@@ -10,6 +10,8 @@ pragma solidity >=0.8.13;
 /// the original calldata must be updated with the new node before forwarding to the actual
 /// resolver logic.
 ///
+/// Reverts `Panic(0x32)` if write is out of bounds.
+///
 library ResolverProfileRewriterLib {
     /// @dev Replace the node in the calldata with a new node.
     ///      Supports `multicall()` to arbitrary depth.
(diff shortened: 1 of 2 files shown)


contracts/src/resolver/libraries/ResolverProfileRewriterLib.sol
```diff
@@ -20,25 +20,47 @@ library ResolverProfileRewriterLib {
...
                     // prettier-ignore
-                    for { } size { size := sub(size, 32) } {
-                        replace(add(add(off, 32), mload(add(off, size))), node)
+                    for { } size { size := sub(size, 32) } { // backwards
+                        let p := mload(add(ptr, size)) // jump[i]
+                        p := add(add(ptr, 32), p) // local ptr
+                        let b := add(p, mload(p)) // local bound w/room for 1 word
+                        if lt(bound, b) {
+                            b := bound // global bound is smaller
+                        }
```
Arachnid, line 48: I don't think we should be clamping here - won't it result in writing to the wrong area of memory? Instead we should skip the replace if the pointer is outside the remaining length of the data.

contracts/src/resolver/libraries/ResolverProfileRewriterLib.sol
```diff
@@ -20,25 +20,47 @@ library ResolverProfileRewriterLib {
...
-                    let size := shl(5, mload(off))
+                    ptr := add(ptr, mload(ptr)) // follow jump
+                    let size := shl(5, mload(ptr)) // read word count as size
                     // prettier-ignore
-                    for { } size { size := sub(size, 32) } {
-                        replace(add(add(off, 32), mload(add(off, size))), node)
+                    for { } size { size := sub(size, 32) } { // backwards
+                        let p := mload(add(ptr, size)) // jump[i]
+                        p := add(add(ptr, 32), p) // local ptr
+                        let b := add(p, mload(p)) // local bound w/room for 1 word
```
Arachnid, line 45: I'm not sure I understand what this comment is trying to convey.

Arachnid (changes requested):

contracts/src/resolver/libraries/ResolverProfileRewriterLib.sol
adraffy: I start with `replace(start, end)` and then recursive calls are `replace(start+offset, min(end, start+offset+size))`

(where `start` is the first byte of raw data: `<length><bytes...>`)
(where `end` is 32 bytes before the actual end: `start + length`)

And any write (the actual node replacement) is avoided if `end < start + 36` aka there aren't 32 bytes of room after the length and selector.

I can take another pass at refactoring it.

contracts/src/resolver/libraries/ResolverProfileRewriterLib.sol
adraffy:
```ts
bytes[] is encoded as: <count> <offset[0]> <offset[1]> ... <offset[count-1]>
each offset[i] points to `bytes` which is `<length> <data...>`

size = count * 32
for loop effectively starts at `offset[count-1]` and goes backwards:
    // p for element pointer
    p = <memory offset of array> + offset[i] // memory offset of bytes[i]
    // b for element bound
    b = p[0] // b is length at first bytes of p
    replace(p, b)
```

I'm keeping the `(pointer, bound)` logic like:
```
bytes data = 'abcde'       end of data
                           v
+--------32--------+--------32--------+
|          .length |<--5-->|    bytes |
|                5 |"abcde"|          |
+------------------+------------------+
|<-----------37----------->|
|     |                    |
^     |<--------32-------->|
ptr   |  room for 1 word
      |
      ^
      bound (ptr + length)
```

But I can refactor it if needed.

contracts/src/resolver/libraries/ResolverProfileRewriterLib.sol
adraffy: Instead of reverting on replacement, I just don't replace, and then let the resolver deal with it.

PR #275, opened 2026-04-17
Add `ETHRegistrar.GRACE_PERIOD`

adraffy:
- added `IPaymentTokenOracle`
     * moved token related things
- changed `IRentPriceOracle`
     * inherits `IPaymentTokenOracle`
     * removed `rentPrice()`
     * added `registerPrice()` and `renewPrice()`
          - priced via components instead of `expiry`
- changed `StandardRentPriceOracle`
     * return zero instead of reverting when possible
     * changed `premiumPrice()` to round down
- changed `ETHRegistrar`
     * inherits `IPaymentTokenOracle` instead of `IRentPriceOracle`
     * added `MIN_RENEW_DURATION`
     * added `GRACE_PERIOD`
     * refactored pricing logic around `rentPrice()`
     * `renew()` during grace pays full price for "debt period"

---

- removed `HCA` from `MockERC20`
- updated deploy scripts

### DIFF
--- a/contracts/deploy/00_MockTokens.ts
+++ b/contracts/deploy/00_MockTokens.ts
@@ -2,25 +2,21 @@ import { artifacts, execute } from "@rocketh";
 
 export default execute(
   async ({ deploy, get, namedAccounts: { deployer } }) => {
-    const hcaFactory =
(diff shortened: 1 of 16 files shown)


adraffy:
#286 is better

PR #283, opened 2026-04-22
Fix `ResolverProfileRewriterLib` memory overflow [Again]

adraffy:
Reopen #274

---

- fixed `ResolverProfileRewriterLib.replaceNode()`
     * avoids writing if out of bounds
- added tests for malicious calldata

---

```ts
bytes[] is encoded as: <count> <offset[0]> <offset[1]> ... <offset[count-1]>
each offset[i] points to `bytes` which is `<length> <data...>`

size = count * 32
for loop effectively starts at `offset[count-1]` and goes backwards:
    // p for element pointer
    p = <memory offset of array> + offset[i] // memory offset of bytes[i]
    // b for element bound
    b = p[0] // b is length at first bytes of p
    replace(p, b)
```

I'm keeping the `(pointer, bound)` logic like:
```
bytes data = 'abcde'       end of data
                           v
+--------32--------+--------32--------+
|          .length |<--5-->|    bytes |
|                5 |"abcde"|          |
+------------------+------------------+
|<-----------37----------->|
|     |                    |
^     |<--------32-------->|
ptr   |  room for 1 word
      |
      ^
      bound (ptr + length)
```

### DIFF
--- a/contracts/src/resolver/libraries/ResolverProfileRewriterLib.sol
+++ b/contracts/src/resolver/libraries/ResolverProfileRewriterLib.sol
@@ -10,6 +10,8 @@ pragma solidity >=0.8.13;
 /// the original calldata must be updated with the new node before forwarding to the actual
 /// resolver logic.
 ///
+/// Reverts `Panic(0x32)` if write is out of bounds.
+///
 library ResolverProfileRewriterLib {
     /// @dev Replace the node in the calldata with a new node.
     ///      Supports `multicall()` to arbitrary depth.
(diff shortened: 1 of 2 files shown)


contracts/src/resolver/libraries/ResolverProfileRewriterLib.sol
```diff
@@ -20,25 +20,47 @@ library ResolverProfileRewriterLib {
...
+                        replace(p, b, node)
                     }
                 }
                 default {
-                    mstore(add(ptr, 36), node) // replace node
+                    // only bound checks on write
+                    if lt(bound, ptr) {
+                        // mstore(0, 0x4e487b71) // error Panic(uint256)
+                        // mstore(32, 0x32) // code
+                        // revert(28, 36) // 32-4, 4+32
```
Arachnid: Should these lines be here?

Arachnid (changes requested):

contracts/src/resolver/libraries/ResolverProfileRewriterLib.sol
adraffy: Removed, was leftover from deciding if it should revert or just not replace.

Arachnid (approved):

PR #287, opened 2026-04-27
Tag migrations onchain

adraffy:
- changed `PermissionedRegistry` to assign token `ROLE_WAS_RESERVED` when registered via reservation
- added tests


Arachnid (changes requested):
I think it's safer to have a new role `ROLE_WAS_RESERVED` and assign that; I know `ROLE_REGISTER_RESERVED` has no per-token meaning but this seems both safer and cleaner, and we're not short of role bits.

TateB:
@codex

TateB (approved):

Arachnid (approved):

TateB (approved):

PR #290, opened 2026-04-29
fix: gate wrapper registry upgrades

TateB:
changes:
- added `ApprovedUpgradeGate` contract, simple allowlist for approved implementations, expectedly DAO-managed.
- added gate check in `WrapperRegistry` upgrade authorisation

note:
`ROLE_UPGRADE` is still used to determine if the user can initiate the upgrade, but the implementation can only be upgraded to implementations specified in the gate contract.

Fixes BET-606

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -2,7 +2,7 @@
 
 - When running tests they should be run from the contracts/ subfolder.
 - See package.json for the test commands.
-- When testing for event emission never use vm.expectEmit, always use vm.recordLogs and actually check the logs properly.
+- When testing for event emission, always check logs properly. Use either `vm.recordLogs` with explicit log assertions, or `vm.expectEmit` paired with an `emit Event(...)` expectation before the call under test.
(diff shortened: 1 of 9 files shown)


Arachnid (approved):

adraffy (commented):
I was expecting:
* 1 gate per type:  `PermReg, PermRes, WrapperReg`
* transition table: `(old, new) => boolean`

Since this only applies to `WrapperRegistry`, is this effectively `WrapperRegistryUpgradeGate`?

TateB:
@adraffy 
permreg/permres don't need it, users can upgrade to an unsafe implementation if they really want to.

transition table is unnecessary since the upgrade hooks from verifiable factory already effectively allow that, you can specify in-implementation which prior implementations can be upgraded from

it is effectively `WrapperRegistryUpgradeGate`, nothing stopping it being reused though

adraffy (approved):

PR #295, opened 2026-05-03
Add Storage Gaps

adraffy:
- added `uint256[256] private __gap` to `EnhancedAccessControl`
- added `uint256[256] private __gap` to `PermissionedRegistry`

---

See linear ticket for more information.


Arachnid (dismissed):
May as well make the gaps bigger, since they cost us nothing - say 256 elements?

TateB (approved):

PR #309, opened 2026-05-16
Add Encoded Label Support to `Graveyard`

adraffy:
- updated `Graveyard.clear()` to process encoded labels using modified encoding

### DIFF
--- a/contracts/src/migration/Graveyard.sol
+++ b/contracts/src/migration/Graveyard.sol
@@ -6,6 +6,7 @@ import {
 } from "@ens/contracts/ethregistrar/BaseRegistrarImplementation.sol";
 import {IBaseRegistrar} from "@ens/contracts/ethregistrar/IBaseRegistrar.sol";
 import {ENS} from "@ens/contracts/registry/ENS.sol";
+import {HexUtils} from "@ens/contracts/utils/HexUtils.sol";
 import {NameCoder} from "@ens/contracts/utils/NameCoder.sol";
 import {INameWrapper} from "@ens/contracts/wrapper/INameWrapper.sol";
 import {ERC1155Holder} from "@openzeppelin/contracts/token/ERC1155/utils/ERC1155Holder.sol";
@@ -72,10 +73,16 @@ contract Graveyard is ERC721Holder, ERC1155Holder {
     ////////////////////////////////////////////////////////////////////////
 
     /// @notice Clear registry for migrated names.
+    /// @dev Append non-zero byte to process encoded-labels.
     /// @param names The array of names to clear.
     function clear(bytes[] calldata names) external {
         for (uint256 i; i < names.length; ++i) {
-            _clear(names[i], 0);
+            bytes calldata name = names[i];
+            bool encoded = name.length >= 1 && uint8(name[name.length - 1]) != 0;
+            if (encoded) {
+                name = name[:name.length - 1]; // remove tail byte
+            }
+            _clear(name, 0, encoded);
         }
     }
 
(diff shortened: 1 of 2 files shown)


contracts/src/migration/Graveyard.sol
```diff
@@ -134,9 +150,10 @@ contract Graveyard is ERC721Holder, ERC1155Holder {
                 // resolver is cleared by migration
                 if (LibMigration.isLocked(fuses)) {
                     return (node, State.LOCKED);
-                } else if (LibMigration.isEmancipatedChild(fuses)) {
-                    return (node, State.OWNED);
                 }
+                // } else if (LibMigration.isEmancipatedChild(fuses)) {
```
hiddentao, line 155: Why is this commented out?

hiddentao (approved):

contracts/src/migration/Graveyard.sol
adraffy: I noticed it was the same as the fallthru.  I left it commented for reference.

contracts/src/migration/Graveyard.sol
```diff
@@ -91,25 +86,23 @@ contract Graveyard is ERC721Holder, ERC1155Holder {
...
-            return (bytes32(0), State.ROOT);
-        }
-        bytes32 parentNode;
-        (parentNode, state) = _clear(name, nextOffset, encoded);
+    function _clear(bytes calldata name, uint256 offset) internal returns (bytes32 node, State) {
         bytes32 labelHash;
-        if (encoded && size == 66 && name[offset + 1] == "[" && name[nextOffset - 1] == "]") {
-            (labelHash, encoded) = HexUtils.hexStringToBytes32(name[offset + 2:offset + 66], 0, 64);
+        uint256 nextOffset;
+        if (offset + 1 < name.length && uint8(name[offset]) == 0) {
```
Arachnid, line 101: This needs a comment explaining what's going on. We also need a way to escape non-normalized names that already had a leading 0-byte - possibly with \x00\x00 for escaping and \x00\x01 for hashes.

adraffy: This is different.  Since `Graveyard` is the only user of this encoding, it's exploiting the fact that a zero-length label in the middle of a name is invalid.   So the only change is that, if it encounters a zero-length label, it treats the next 32 bytes as the hash.

Recap:
* wrapped names are all revealed by definition
* unwrapped names could have no pre-image or be a literal encoded label

Using `byte(0) + bytes32(labelHash)` in the middle of a name gives per-label control on interpreting the label as any preimage or any hash.

I can add some comments.

contracts/src/migration/Graveyard.sol
```diff
@@ -84,16 +86,23 @@ contract Graveyard is ERC721Holder, ERC1155Holder {
...
-    {
-        (bytes32 labelHash, uint256 nextOffset) = NameCoder.readLabel(name, offset);
-        if (labelHash == bytes32(0)) {
-            return (bytes32(0), State.ROOT);
+    function _clear(bytes calldata name, uint256 offset) internal returns (bytes32 node, State) {
+        bytes32 labelHash;
+        uint256 nextOffset;
+        if (offset + 1 < name.length && uint8(name[offset]) == 0) {
+            ++offset;
+            nextOffset = offset + 32;
```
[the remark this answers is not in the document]
adraffy: all wrapped names are revealed

the caller of this contract would only use encoded labels for unwrapped names

contracts/src/migration/Graveyard.sol
adraffy: Added comments

adraffy: This should support clearing the following labels:
* `""` &mdash; empty label 
* `bytes32(0)` &mdash; zero labelhash
* `[0000000000000000000000000000000000000000000000000000000000000000]` &mdash; literal labelhash

contracts/src/migration/Graveyard.sol
TateB: remove?

TateB (approved):

PR #321, opened 2026-05-25
[Idea 1] fix: when a locked wrapped name is migrated, then transferred to a new owner, the new owner does not get ownership of the subregistry

hiddentao:


hiddentao:
> ### 💡 Codex Review
> https://github.com/ensdomains/contracts-v2/blob/ed94592cf1ff26d5a4b9253cbc35fba5d0762f63/contracts/src/migration/LockedWrapperReceiver.sol#L178-L179
> 
> 
> **P1 Badge Migrate emancipated children without fuse-blocked resolver writes**
> In the `isEmancipatedChild` migration path, calling `NAME_WRAPPER.setResolver(node, address(0))` before `unwrap` causes migration to revert for children that have the `CANNOT_SET_RESOLVER` fuse burned (a valid emancipated state). This makes those wrapped children unmigratable even though the previous implementation handled them by unwrapping first and then clearing the resolver directly in ENSv1. The regression affects any detached child migrated through `LockedWrapperReceiver._migrateWrapped` where resolver updates are fuse-locked.
> 
> ℹ️ About Codex in GitHub

Not an issue because the `CANNOT_SET_RESOLVER` can only be burned if the name is locked, meaning this if-condition branch would never get executed anyway.

adraffy:
Went with #331

PR #326, opened 2026-05-26
[Idea 3] Add `WrapperRegistry.reclaim()`

adraffy:
- added `RegistryRolesLib.ROLE_WRAPPER_RECLAIM`
- changed `WrapperRegistry` to have at most one account with admin
- added `WrapperRegistry.reclaim(from, to)`
     * requires `ROLE_WRAPPER_RECLAIM` on token for caller
     * requires `ROLE_WRAPPER_RECLAIM` on root for `from`

### DIFF
--- a/contracts/src/access-control/libraries/EACBaseRolesLib.sol
+++ b/contracts/src/access-control/libraries/EACBaseRolesLib.sol
@@ -11,6 +11,10 @@ pragma solidity ^0.8.20;
 /// just the 32 admin role slots. Used to extract which admin roles an account holds.
 ///
 library EACBaseRolesLib {
+    ////////////////////////////////////////////////////////////////////////
+    // Constants
+    ////////////////////////////////////////////////////////////////////////
+
     /// @dev Mask with bit 0 set in every nybble — represents one unit per role slot across all 64 slots.
     uint256 internal constant ALL_ROLES =
         0x1111111111111111111111111111111111111111111111111111111111111111;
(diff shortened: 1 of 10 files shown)


contracts/src/access-control/libraries/EACBaseRolesLib.sol
```diff
@@ -11,11 +11,28 @@ pragma solidity ^0.8.20;
...
     /// @dev Mask selecting only the 32 admin role nybbles (upper 128 bits).
     uint256 internal constant ADMIN_ROLES =
         0x1111111111111111111111111111111100000000000000000000000000000000;
+
+    ////////////////////////////////////////////////////////////////////////
+    // Implementation
+    ////////////////////////////////////////////////////////////////////////
+
+    /// @dev Derive roles bitmap from role counts for a resource.
+    function fromCounts(uint256 count) internal pure returns (uint256) {
```
Arachnid: This won't return consistent results - it guarantees the least significant bit of each range is set, but the other three bits are in an inconsistent state depending on the previous counts. It needs to be masked after the bitwise ops to give a correct result.

Arachnid (changes requested):

contracts/src/access-control/libraries/EACBaseRolesLib.sol
adraffy: simplified

Arachnid (approved):

contracts/src/registry/WrapperRegistry.sol
```diff
@@ -144,6 +145,30 @@ contract WrapperRegistry is
...
+                RegistryRolesLib.ROLE_WRAPPER_RECLAIM,
+                sender
+            );
+        }
+        (uint256 counts, ) = getAssigneeCount(ROOT_RESOURCE, EACBaseRolesLib.ALL_ROLES);
+        uint256 unionRoles = EACBaseRolesLib.fromCounts(counts);
+        for (uint256 i; i < revokes.length; ++i) {
+            _revokeRoles(ROOT_RESOURCE, EACBaseRolesLib.ALL_ROLES, revokes[i], false);
+        }
+        _grantRoles(ROOT_RESOURCE, EACBaseRolesLib.withAdminRolesApplied(unionRoles), to, false);
```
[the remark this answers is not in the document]
adraffy: agreed, changed

adraffy:
Went with #331

PR #331, opened 2026-06-01
[Idea 4] Change `WrapperRegistry` to use virtual root admin

adraffy:
- changed `WrapperRegistry` to have at most one account with admin
    * roles granted to "virtual owner" — actual owner is parent registry
    * `_msgSender()` remaps from token owner to "virtual owner"
- changed `WrapperRegistry` to respect frozen `NameWrapper` fuses for `ROLE_RENEW`, `ROLE_UPGRADE`, and `ROLE_CAN_NAME` (like `ROLE_REGISTRAR`)

### DIFF
--- a/contracts/src/migration/LockedWrapperReceiver.sol
+++ b/contracts/src/migration/LockedWrapperReceiver.sol
@@ -242,7 +242,10 @@ abstract contract LockedWrapperReceiver is AbstractWrapperReceiver {
             roleBitmap |= roleBitmap << 128; // give admin
         }
         if ((fuses & CANNOT_TRANSFER) == 0) {
-            roleBitmap |= RegistryRolesLib.ROLE_CAN_TRANSFER_ADMIN;
+            roleBitmap |=
+                RegistryRolesLib.ROLE_CAN_TRANSFER_ADMIN |
+                RegistryRolesLib.ROLE_WRAPPER_RECLAIM |
+                RegistryRolesLib.ROLE_WRAPPER_RECLAIM_ADMIN;
         }
     }
 }
--- a/contracts/src/migration/LockedWrapperReceiver.sol
+++ b/contracts/src/migration/LockedWrapperReceiver.sol
@@ -155,7 +155,6 @@ abstract contract LockedWrapperReceiver is AbstractWrapperReceiver {
                                     node,
                                     parentRegistry,
                                     md.label,
-                                    md.owner,
                                     _subregistryRoleBitmapFromFuses(fuses)
                                 )
                             )
@@ -225,7 +224,6 @@ abstract contract LockedWrapperReceiver is AbstractWrapperReceiver {
         if (LibMigration.notFrozen(fuses)) {
             roleBitmap |= roleBitmap << 128; // give admin
         }
-        roleBitmap |= RegistryRolesLib.ROLE_WRAPPER_RECLAIM; // no admin, used for "tracking"
     }
 
     /// @dev Convert fuses to equivalent token roles.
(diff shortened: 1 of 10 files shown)


contracts/src/registry/WrapperRegistry.sol
```diff
@@ -233,6 +223,33 @@ contract WrapperRegistry is
...
+    /// @param account The account to get settable roles for.
+    /// @return The settable roles.
+    function _getSettableRoles(uint256 resource, address account)
+        internal
+        view
+        override
+        returns (uint256)
+    {
+        uint256 roleBitmap = super._getSettableRoles(resource, account);
+        return resource == ROOT_RESOURCE ? roleBitmap >> 128 : roleBitmap;
```
Arachnid, line 262: `_getSettableRoles` already provides user roles, so masking admin roles out here would be cleaner than shifting them.

contracts/src/registry/WrapperRegistry.sol
```diff
@@ -233,6 +223,33 @@ contract WrapperRegistry is
         return _register(label, owner, subregistry, resolver, roleBitmap, expiry, false);
     }
 
+    /// @dev If the caller is the token owner, return the virtual owner. 
+    function _msgSender() internal view override returns (address) {
+        address sender = super._msgSender();
+        return
+            sender == PermissionedRegistry(address(_parentRegistry)).findOwner(_childLabel)
+                ? address(this)
+                : sender;
```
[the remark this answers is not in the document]
Arachnid: Agree with this - the override should only apply to EAC, if possible.

contracts/src/migration/LockedWrapperReceiver.sol
```diff
@@ -218,16 +217,13 @@ abstract contract LockedWrapperReceiver is AbstractWrapperReceiver {
...
             RegistryRolesLib.ROLE_RENEW |
-            RegistryRolesLib.ROLE_RENEW_ADMIN |
             RegistryRolesLib.ROLE_UPGRADE |
-            RegistryRolesLib.ROLE_UPGRADE_ADMIN |
-            RegistryRolesLib.ROLE_CAN_NAME |
-            RegistryRolesLib.ROLE_CAN_NAME_ADMIN;
+            RegistryRolesLib.ROLE_CAN_NAME;
+        if (LibMigration.notFrozen(fuses)) {
+            roleBitmap |= roleBitmap << 128; // give admin
+        }
```
Arachnid, line 226: What's the reasoning behind this change?

Arachnid (changes requested):

contracts/src/migration/LockedWrapperReceiver.sol
adraffy: By moving above the `notFrozen` logic, admin only get granted if the wrapper wasn't frozen, similar to how `ROLE_REGISTRAR` works.

We talked about it a bit yesterday.  Codex brought it up as a potential issue [here](https://github.com/ensdomains/contracts-v2/pull/326#pullrequestreview-4402833448).  Both ways seem okay to me, since you disagree I can revert it.

Arachnid: I don't disagree - this seems fine.

Arachnid (approved):

contracts/src/registry/WrapperRegistry.sol
```diff
@@ -137,8 +127,9 @@ contract WrapperRegistry is
         _parentRegistry = parentRegistry;
         _childLabel = childLabel;
         emit RegistryCreated();
-        emit ParentUpdated(parentRegistry, childLabel, rootAccount);
-        _grantRoles(ROOT_RESOURCE, roleBitmap, rootAccount, false);
+        address virtualOwner = address(_parentRegistry);
+        emit ParentUpdated(parentRegistry, childLabel, virtualOwner);
+        _grantRoles(ROOT_RESOURCE, roleBitmap, virtualOwner, false);
```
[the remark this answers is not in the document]
adraffy: I think this is fixable.

contracts/src/registry/WrapperRegistry.sol
adraffy: I think this has to stay as-is.  Essentially, I need to [block the parent logic](https://github.com/ensdomains/contracts-v2/blob/post-audit/contracts/src/registry/PermissionedRegistry.sol#L567-L580) for root, and make it so only the user bits that have admin are settable.

Arachnid (approved):

PR #333, opened 2026-06-03
Leverage `PermissionRegistry.setApprovalForAll()` in `EAC`

adraffy:
- changed `EnhancedAccessControl`
    * added virtual `_getRoles(resource, account): roles` which permits logic <ins>before</ins> permission checks
    * added private `_effectiveRoles(resource, account)` which is root + resource
    * refactored to utilize those functions
- changed `PermissionedRegistry`
    * overrode `_getRoles()` to account for token expiry and approval 
    * simplified `_getSettableRoles()` and `_getRevokableRoles()` since `_checkedRoles()` already handles expiry
- changed `WrapperRegistry`
    * removed `_msgSender()` hack
    * removed `IContractNamer` override
    * overrode `_getRoles()` to account for virtual owner — applies <ins>after</ins> expiry and approval checks, so approvals are not virtualized and do not survive transfer

---

* `EAC` reverts use the caller address
* `PermissionRegistry.roles()` returns the same roles for the token owner AND virtual owner AND approved
* `test_revokeRoles_whileReserved()` returns false instead of reverting
* `test_revokeRoles_whileExpired()` returns false instead of reverting

### DIFF
--- a/contracts/src/access-control/EnhancedAccessControl.sol
+++ b/contracts/src/access-control/EnhancedAccessControl.sol
@@ -170,7 +170,7 @@ abstract contract EnhancedAccessControl is HCAContext, ERC165, IEnhancedAccessCo
 
     /// @inheritdoc IEnhancedAccessControl
     function roles(uint256 resource, address account) public view virtual returns (uint256) {
-        return _roles[resource][account];
+        return _getRoles(resource, account);
     }
 
     /// @inheritdoc IEnhancedAccessControl
@@ -180,7 +180,7 @@ abstract contract EnhancedAccessControl is HCAContext, ERC165, IEnhancedAccessCo
 
     /// @inheritdoc IEnhancedAccessControl
     function hasRootRoles(uint256 roleBitmap, address account) public view virtual returns (bool) {
-        return _roles[ROOT_RESOURCE][account] & roleBitmap == roleBitmap;
+        return _getRoles(ROOT_RESOURCE, account) & roleBitmap == roleBitmap;
     }
 
     /// @inheritdoc IEnhancedAccessControl
@@ -191,7 +191,8 @@ abstract contract EnhancedAccessControl is HCAContext, ERC165, IEnhancedAccessCo
         returns (bool)
     {
         return
-            (_roles[ROOT_RESOURCE][account] | _roles[resource][account]) & roleBitmap == roleBitmap;
+            (_getRoles(ROOT_RESOURCE, account) | _getRoles(resource, account)) & roleBitmap ==
+            roleBitmap;
     }
 
     /// @inheritdoc IEnhancedAccessControl
@@ -424,6 +425,7 @@ abstract contract EnhancedAccessControl is HCAContext, ERC165, IEnhancedAccessCo
         virtual
         returns (uint256)
     {
+        account = _getUnderlyingAccount(resource, account);
         return
             EACBaseRolesLib.withAdminRolesApplied(
                 _roles[resource][account] | _roles[ROOT_RESOURCE][account]
@@ -443,15 +445,31 @@ abstract contract EnhancedAccessControl is HCAContext, ERC165, IEnhancedAccessCo
         virtual
         returns (uint256)
     {
+        account = _getUnderlyingAccount(resource, account);
         return
             EACBaseRolesLib.withAdminRolesApplied(
                 _roles[resource][account] | _roles[ROOT_RESOURCE][account]
             );
     }
 
+    /// @dev Returns the underlying account with permissions.
+    function _getUnderlyingAccount(uint256 resource, address operator)
+        internal
+        view
+        virtual
+        returns (address)
+    {
+        return operator;
+    }
+
     ////////////////////////////////////////////////////////////////////////
     // Private Functions
     ////////////////////////////////////////////////////////////////////////
+
+    /// @dev Returns the roles bitmap for an account in a resource.
+    function _getRoles(uint256 resource, address account) private view returns (uint256) {
+        return _roles[resource][_getUnderlyingAccount(resource, account)];
+    }
 
     /// @dev Checks if a role bitmap contains only valid role bits.
     /// @param roleBitmap The role bitmap to check.
--- a/contracts/src/access-control/EnhancedAccessControl.sol
+++ b/contracts/src/access-control/EnhancedAccessControl.sol
@@ -190,9 +190,7 @@ abstract contract EnhancedAccessControl is HCAContext, ERC165, IEnhancedAccessCo
         virtual
         returns (bool)
     {
-        return
-            (_getRoles(ROOT_RESOURCE, account) | _getRoles(resource, account)) & roleBitmap ==
-            roleBitmap;
+        return _getEffectiveRoles(resource, account) & roleBitmap == roleBitmap;
     }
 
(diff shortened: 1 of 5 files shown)


contracts/src/access-control/EnhancedAccessControl.sol
```diff
@@ -443,16 +440,33 @@ abstract contract EnhancedAccessControl is HCAContext, ERC165, IEnhancedAccessCo
...
     ////////////////////////////////////////////////////////////////////////
 
+    /// @dev Returns the roles bitmap for an account in a resource.
+    function _getRoles(uint256 resource, address account) private view returns (uint256) {
+        return _roles[resource][_getUnderlyingAccount(resource, account)];
+    }
+
+    /// @dev Returns the effective roles bitmap for an account.
+    function _getEffectiveRoles(uint256 resource, address account) private view returns (uint256) {
+        return _getRoles(ROOT_RESOURCE, account) | _getRoles(resource, account);
```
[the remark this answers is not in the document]
adraffy: `setApprovalForAll()` is only for tokens and ignored for root

contracts/src/registry/PermissionedRegistry.sol
```diff
@@ -572,34 +572,29 @@ contract PermissionedRegistry is ERC1155Singleton, EnhancedAccessControl, IPermi
         returns (uint256)
     {
         uint256 roleBitmap = super._getSettableRoles(resource, account);
-        if (resource == ROOT_RESOURCE) {
-            return roleBitmap;
-        } else if (ownerOf(_constructTokenId(resource, _entry(resource))) == address(0)) {
-            return 0; // available or reserved
-        }
-        return roleBitmap >> 128; // remove admin
+        return resource == ROOT_RESOURCE ? roleBitmap : roleBitmap >> 128;
```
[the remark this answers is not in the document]
adraffy: no, `_getSettableRoles()` uses `_effectiveRoles()` which uses `_checkedRoles()` which respects expired

contracts/src/access-control/EnhancedAccessControl.sol
```diff
@@ -443,16 +439,28 @@ abstract contract EnhancedAccessControl is HCAContext, ERC165, IEnhancedAccessCo
...
     {
-        return
-            EACBaseRolesLib.withAdminRolesApplied(
-                _roles[resource][account] | _roles[ROOT_RESOURCE][account]
-            );
+        return EACBaseRolesLib.withAdminRolesApplied(_effectiveRoles(resource, account));
+    }
+
+    /// @dev Returns the effective roles bitmap for an account.
+    function _checkedRoles(uint256 resource, address account)
```
Arachnid: Could this just be `_getRoles`?

contracts/src/registry/PermissionedRegistry.sol
```diff
@@ -572,34 +572,29 @@ contract PermissionedRegistry is ERC1155Singleton, EnhancedAccessControl, IPermi
...
     }
 
     /// @inheritdoc EnhancedAccessControl
     /// @dev Override for token-dependent logic:
     ///
-    /// Token roles can only be revoked from registered tokens.
-    ///
-    /// Root roles are unaffected.
+    /// * if token is expired (available or reserved), return null.
+    /// * if caller is approved by token, return token owner.
```
Arachnid:
```suggestion
    /// * if caller is approved by token owner, OR the caller's roles with the owner's roles
```

contracts/src/registry/PermissionedRegistry.sol
```diff
@@ -572,34 +572,29 @@ contract PermissionedRegistry is ERC1155Singleton, EnhancedAccessControl, IPermi
...
     ///
-    function _getRevokableRoles(uint256 resource, address account)
+    function _checkedRoles(uint256 resource, address operator)
```
Arachnid: This should be `account`, not `operator`, I think.

contracts/src/registry/WrapperRegistry.sol
```diff
@@ -262,6 +243,29 @@ contract WrapperRegistry is
         return resource == ROOT_RESOURCE ? roleBitmap >> 128 : roleBitmap;
     }
 
+    /// @inheritdoc PermissionedRegistry
+    /// @dev Override for token-dependent logic:
+    ///
+    /// * if root and operator is token owner, remap to virtual owner.
+    ///
+    function _checkedRoles(uint256 resource, address operator)
```
Arachnid: As above, I think this should be called `account`, not `operator`.

contracts/src/registry/WrapperRegistry.sol
```diff
@@ -262,6 +243,29 @@ contract WrapperRegistry is
...
+        returns (uint256 roleBitmap)
+    {
+        roleBitmap = super._checkedRoles(resource, operator);
+        if (resource == ROOT_RESOURCE) {
+            address parent = address(_parentRegistry); // virtual owner
+            if (
+                parent != address(0) &&
+                operator == PermissionedRegistry(parent).findOwner(_childLabel)
+            ) {
+                roleBitmap |= super._checkedRoles(resource, parent);
```
Arachnid: I think in this instance we want to replace the roles rather than OR them; otherwise we have an unintuitive second source of roles. I think this is quantitatively different to the case with `PermissionedResolver` because in that case both accounts obviously exist, whereas here ORing them gives the 'virtual' account an external surface area that didn't previously exist.

Arachnid (changes requested):

contracts/src/access-control/EnhancedAccessControl.sol
adraffy: ya, changed

contracts/src/registry/PermissionedRegistry.sol
adraffy: agreed, changed

contracts/src/registry/WrapperRegistry.sol
adraffy: changed

contracts/src/registry/WrapperRegistry.sol
adraffy: yeah that makes sense, changed and added comment

contracts/src/registry/PermissionedRegistry.sol
```diff
@@ -572,34 +572,29 @@ contract PermissionedRegistry is ERC1155Singleton, EnhancedAccessControl, IPermi
...
-    /// Root roles are unaffected.
+    /// * if token is expired (available or reserved), return null.
+    /// * if caller is approved by token owner, OR the caller's roles with the owner's roles
     ///
-    function _getRevokableRoles(uint256 resource, address account)
+    function _checkedRoles(uint256 resource, address operator)
         internal
         view
+        virtual
         override
```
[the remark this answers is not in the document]
adraffy: mistake

contracts/src/registry/PermissionedRegistry.sol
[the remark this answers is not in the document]
adraffy: incorrect, as explained before `_getSettableRoles` invokes `_getRoles` which does expiry check (it's just hard to follow through the different levels of inheritance)

Arachnid (approved):

TateB (approved):

PR #340, opened 2026-06-09
Add `AbstractETHRegistrar.renewBatch()`

adraffy:
- added `renewBatch(labels[], duration, isExpiry, paymentToken, referrer)` and tests


adraffy:
use HCA

PR #350, opened 2026-06-17
Make `Graveyard.clear()` clear TTL for expired locked names

adraffy:
- changed conditional `setResolver()` to `setRecord()`
- changed tests to check TTL

### DIFF
--- a/contracts/test/unit/migration/Graveyard.t.sol
+++ b/contracts/test/unit/migration/Graveyard.t.sol
@@ -113,7 +113,7 @@ contract GraveyardTest is MigrationControllerFixture {
         graveyard.clear(_oneName(abi.encodePacked(uint8(0), labelHash, name))); // mark as prehashed
         assertEq(registryV1.resolver(node), address(1), "uncleared");
         graveyard.clear(_oneName(NameCoder.addLabel(name, label)));
-        assertEq(registryV1.resolver(node), address(0), "after");
+        _checkCleared(node, "after");
     }
 
     function test_clear_prehashedLabel_hashed(bytes32 labelHash) external {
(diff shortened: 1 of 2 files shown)


contracts/test/unit/migration/Graveyard.t.sol
```diff
@@ -414,15 +410,25 @@ contract GraveyardTest is MigrationControllerFixture {
...
     ////////////////////////////////////////////////////////////////////////
 
+    function _checkCleared(bytes memory name) internal {
+        _checkCleared(NameCoder.namehash(name, 0), NameCoder.decode(name));
+    }
+
+    /// @dev Ensure resolver and TTL are cleared.
+    function _checkCleared(bytes32 node, string memory tag) internal {
+        assertEq(registryV1.resolver(node), address(0), string.concat("resolver:", tag));
+        assertEq(registryV1.ttl(node), 0, string.concat("ttl:", tag));
```
makoto, line 429: Do we have a test scenario with non 0 ttl?

makoto (changes requested):

contracts/test/unit/migration/Graveyard.t.sol
adraffy: good idea, adding

adraffy:
Converting to draft as I now think this isn't necessary.

---

Migration flow only clears the resolver.  Using `set(Subnode)Record()` to clear resolver and TTL together emits an extra `NewOwner` event.  IMO it's a waste of gas to check and clear TTL using that approach or separately via `setTTL()`.   The graveyard could clear it for descendants of migrated children (what this PR currently does) but the migration controllers need modified to clear it for migrated tokens.

Since TTL has no use in the protocol, leave it as-is.

contracts/test/unit/migration/Graveyard.t.sol
adraffy: didn't implement: decided against merging

PR #354, opened 2026-06-20
Refactored `PermissionedResolver` (inodes w/o versioning)

adraffy:
* added name-based setters
* ~~`SharedResolver` is a `PublicResolver` that respects the ENSv2 registry and uses name-based setters.~~
* `PermissionedResolver` now uses "inodes" instead of versioning and uses name-based setters.

---

- added `IRecordResolver`
     * designed around `event Linked(recordId, node, name)`
- added `resolvers/interfaces/setters/` with `name`-based setters
- added `AbstractRecordResolver` as standard core implementation
- refactored `PermissionedResolver` and tests
    * removed versioning
    * removed per-name permissions
- changed `LibRegistry`
    * changed `findOwner()` to `findExactOwner()`
    * added `findNearestOwner()`
    * added `findNearestRegistry()`
- updated `UniversalHelper` to match
 
---

- changed `utils.ts`
     * `namehash()` and `labelhash()` do not use encoded labels
- changed `resolutions.ts`
     * changed `write` to `writeV1`
     * added `writeV2`


contracts/test/unit/resolver/PermissionedResolver.t.sol
```diff
@@ -156,726 +170,808 @@ contract PermissionedResolverTest is Test {
...
+    ////////////////////////////////////////////////////////////////////////
+    // EAC grantRoles() is disabled
+    ////////////////////////////////////////////////////////////////////////
+
+    function test_grantRoles_disabled(uint256 resource, uint8 roleBit, address account) external {
+        vm.assume(resource > 0 && roleBit < 64);
+        uint256 roleBitmap = 1 << (roleBit << 2);
+
+        vm.prank(owner);
         vm.expectRevert(
```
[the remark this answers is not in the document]
adraffy: fixed

contracts/src/resolver/interfaces/IPermissionedResolver.sol
```diff
@@ -1,61 +1,92 @@
...
+    /// @notice Authorize fine-grained permission to `account`.
+    /// @param setter The ABI-encoded setter calldata to authorize.
+    /// @param account The account to be authorize for role.
+    /// @return `true` if an authorization was changed.
+    function grantSetterRoles(bytes calldata setter, address account) external returns (bool);
+
+    /// @notice Associate `name` with `targetNode`.
+    /// @param sourceName The DNS-encoded name to link.
+    /// @param targetNode The target namehash or null to unlink.
+    function link(bytes calldata sourceName, bytes32 targetNode) external;
```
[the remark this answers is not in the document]
adraffy: fixed

contracts/src/resolver/SharedResolver.sol
```diff
@@ -0,0 +1,219 @@
...
+    }
+
+    /// @inheritdoc AbstractRecordResolver
+    function supportsInterface(bytes4 interfaceId)
+        public
+        view
+        override(AbstractRecordResolver, DelegatedContractNamer)
+        returns (bool)
+    {
+        return super.supportsInterface(interfaceId);
```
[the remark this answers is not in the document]
adraffy: fixed

contracts/src/resolver/PermissionedResolver.sol
```diff
@@ -749,8 +363,17 @@ contract PermissionedResolver is
...
+    {}
+
+    /// @dev Ensure record exists for `name`.
+    function _ensureRecord(bytes memory name) internal returns (uint256 recordId) {
+        bytes32 node = NameCoder.namehash(name, 0);
+        recordId = _recordIds[node];
+        if (recordId == 0) {
+            recordId = ++_recordCount;
+            _recordIds[node] = recordId;
+            emit Linked(node, name, recordId);
```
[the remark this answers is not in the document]
adraffy: I think this is best as-is

contracts/src/resolver/PermissionedResolver.sol
```diff
@@ -1,190 +1,100 @@
...
+///
+/// | Function           | Role                   | Argument            |
+/// | ------------------ | ---------------------- | ------------------- |
+/// | `setABI()`         | `ROLE_SET_ABI`         | uint256 contentType |
+/// | `setAddress()`     | `ROLE_SET_ADDRESS`     | uint256 coinType    |
+/// | `setContentHash()` | `ROLE_SET_CONTENTHASH` |                     |
+/// | `setData()`        | `ROLE_SET_DATA`        | string key          |
+/// | `setInterface()`   | `ROLE_SET_INTERFACE`   | bytes4 interfaceId  |
+/// | `setName()`        | `ROLE_SET_NAME`        |                     |
+/// | `setText()`        | `ROLE_SET_TEXT`        | string key          |
```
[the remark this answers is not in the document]
adraffy: oops, fixed

contracts/src/resolver/interfaces/IPermissionedResolver.sol
```diff
@@ -1,61 +1,60 @@
...
+
+    /// @notice Associate `name` with `targetNode`.
+    /// @param sourceName The DNS-encoded name to link.
+    /// @param targetNode The target namehash.
+    function linkToNode(bytes calldata sourceName, bytes32 targetNode) external;
+
+    /// @notice Associate `name` with `recordId`.
+    /// @param sourceName The DNS-encoded name to link.
+    /// @param recordId The record ID or 0 to unlink.
+    function linkToRecord(bytes calldata sourceName, uint256 recordId) external;
```
Arachnid, line 41: Do we need this? Filesystems, for example, don't tend to expose this.

contracts/src/resolver/interfaces/IPermissionedResolver.sol
```diff
@@ -1,61 +1,60 @@
...
+
+    /// @notice Get the number of created records.
+    function getRecordCount() external view returns (uint256);
+
+    /// @notice Get the record linked to `node`.
+    /// @param node The namehash to find.
+    /// @return The record ID or 0 if not linked.
+    function getRecordId(bytes32 node) external view returns (uint256);
```
Arachnid, line 49: Likewise - do we need this, or can we keep record details largely internal for onchain callers?

contracts/src/resolver/interfaces/IRecordResolver.sol
```diff
@@ -0,0 +1,55 @@
...
+    /// @notice `name` was associated with a record.
+    ///         If `recordId = 0`, the `name` was unassociated.
+    /// @param recordId The record ID.
+    /// @param node The namehash of name.
+    /// @param name The DNS-encoded name.
+    event Linked(uint256 indexed recordId, bytes32 indexed node, bytes name);
+
+    /// @notice All values assocated with `recordId` were cleared.
+    /// @param recordId The record ID.
+    event Cleared(uint256 indexed recordId);
```
Arachnid, line 42: This is documented as taking a `recordId`, but it's called with a `node`. I believe the latter is correct?

contracts/src/resolver/libraries/PermissionedResolverLib.sol
```diff
@@ -31,68 +36,59 @@ library PermissionedResolverLib {
...
-    /// @return part The computed record-type identifier.
-    function partHash(string memory x) internal pure returns (bytes32) {
-        return keccak256(bytes(x));
+    /// @dev Compute EAC resource from `bytes4` argument.
+    ///      Equivalent to `uint256(keccak256(abi.encodePacked(x)))`.
+    function resource(bytes4 x) internal pure returns (uint256 ret) {
+        assembly {
+            mstore(0, x)
+            ret := keccak256(0, 4)
+        }
```
Arachnid, line 92: Are these being used safely? If we don't include both identifier and parameter, we'll get the same resource ID for different setters.

contracts/src/resolver/PermissionedResolver.sol
```diff
@@ -1,190 +1,100 @@
...
 ///
-/// eg. `setAlias("a.eth", "b.eth")`
-/// * `getAlias("a.eth") => "b.eth"`
-/// * `getAlias("[sub].a.eth") => "[sub].b.eth"`
-/// * `getAlias("[x.y].a.eth") => "[x.y].b.eth"`
-/// * `getAlias("abc.eth") => ""`
+/// `linkToNode(name, node)` makes `name` use the record currently used by `node`.
+/// `linkToRecord(name, recordId)` makes `name` use a specific record ID.
+/// `linkToRecord(name, 0)` unlinks `name` from the record.
+/// `clear(name)` makes `name` use a new record.
```
Arachnid, line 47: Is there a practical difference between these two? If I `linkToRecord(name, 0)` then write to it, does it update record 0?

contracts/src/resolver/PermissionedResolver.sol
```diff
@@ -1,190 +1,100 @@
...
-/// * `setText(key)` can be permissioned with `authorizeTextRoles()`
-///    - caller requires `ROLE_SET_TEXT_ADMIN` on `resource(<namehash>, 0)`
-///    - `ROLE_SET_TEXT` is authorized on `resource(<namehash>, <part>)`
-/// * `setData(key)` can be permissioned with `authorizeDataRoles()`
-///    - caller requires `ROLE_SET_DATA_ADMIN` on `resource(<namehash>, 0)`
-///    - `ROLE_SET_DATA` is authorized on `resource(<namehash>, <part>)`
-/// * `setAddr(coinType)` can be permissioned with `authorizeAddrRoles()`
-///    - caller requires `ROLE_SET_ADDR_ADMIN` on `resource(<namehash>, 0)`
-///    - `ROLE_SET_ADDR` is authorized on `resource(<namehash>, <part>)`
+/// Names without a record use the default record, which can be managed using the root name (`0x00`).
```
Arachnid, line 51: Is this true? It looks like names without a record get assigned one on first write, which seems like the correct pattern.

contracts/src/resolver/SharedResolver.sol
```diff
@@ -0,0 +1,220 @@
...
+    }
+
+    /// @dev Ensure `Linked` is emit once.
+    function _ensureLinked(bytes32 node, bytes calldata name) internal {
+        RecordPointer storage p = _pointers[node];
+        if (p.version == 0) {
+            p.version = 1;
+            emit Linked(uint256(node), node, name);
+        }
+    }
```
Arachnid, line 190: Since we don't allow relinking, couldn't we simply set every node's record ID as its node ID, and emit the event on first write?

Arachnid (changes requested):

contracts/src/resolver/interfaces/IPermissionedResolver.sol
adraffy: Need, no.  Originally link was just `link() = linkToNode()` but then I realized that `bytes32(0)` has 2 interpretations: unlink and the default record.  Instead of adding `unlink()`, I added the more powerful `linkToRecord()` which can link to a previously unreachable record.

contracts/src/resolver/interfaces/IRecordResolver.sol
adraffy: `PermissionedResolver` can't clear using this event because it's records aren't versioned.  Instead, it clears by unlinking and then reassigning a new record.  The earlier version of the `PermissionedResolver` with inodes did support `clear()` but it required internal versioning to accomplish it (which you disliked.)

`SharedResolver` can clear because it's records are versioned.  It also has the property that `recordId == node`.

I kept this function on the `IRecordResolver` interface since it's the common logic for an indexer to reconstruct the resolver state.

contracts/src/resolver/libraries/PermissionedResolverLib.sol
adraffy: In an earlier version, I tagged each hash with a prefix, but it's not needed.   The setter roles are stored per EAC bit.  So when tasked with explaining a `resource` for a role, eg. `ROLE_SET_ADDRESS`, you invert `uint256 resource => bytes arg` from `event ResourceArgument(...)`, and then interpret `arg` based on `ROLE_SET_ADDRESS` which would be `decode(arg, (uint256))`.  The only requirement is that `resource != 0`.

It doesn't matter if multiple role bits use the same resource, since their permissions are separated by the EAC bits.

contracts/src/resolver/PermissionedResolver.sol
adraffy: It's a minor technical one:
* If you only `unlink()`, the name will resolve to default record (if one is defined).
* If you `unlink()` and then `link()` to a new empty record, it's effectively "cleared".

The default record is queried on `PermissionedResolver.resolve()` when a requested node that has no link.  The default record corresponds to the empty/root name.  This gives it `DedicatedResolver`-like behavior and allows intercepting all unlinked queries that hit the resolver.

contracts/src/resolver/SharedResolver.sol
adraffy: That was my thinking and this was my solution for emitting the event once:  if you interact with a record on the `SharedResolver` and the version is 0, it's "first write", and the preimage needs revealed.

Specifically, `_ensureLinked()` was added since you can `approve()` an account on a name before its written to, which wouldn't be invertible unless this also was considered a "first write".

Both resolvers have the property that `event Linked()` is emit before `recordId` is referenced.

contracts/src/resolver/PermissionedResolver.sol
adraffy: `PermissionedResolver` has a default record, since you own it.

`SharedResolver` doesn't have a default record since the owner depends on registry.

contracts/src/resolver/interfaces/IPermissionedResolver.sol
adraffy: Same as above:  I'm okay with removing `getRecordId()` and `linkToRecord()` but would want `unlink(node/node)` to replace it (and change `linkToNode()` back to `link()`).

However, I prefer having this functionality, as it lets you recover old records and `getRecordId()` is just an onchain version of what the indexer would be able to do.

contracts/src/resolver/interfaces/IPermissionedResolver.sol
adraffy: I think my other idea for disambiguating was `link(bytes, bytes)` which would use `""` to clear and `dnsEncode("") = "0x00"` for default name.

contracts/script/testNames/display.ts
```diff
@@ -103,35 +100,33 @@ export async function showName(env: DevnetEnvironment, names: string[]) {
...
-    }
-  }
-
-  if (aliasData.length > 0) {
-    console.log(`\nAlias Information:`);
-    console.table(aliasData);
-  } else {
-    console.log(`\nNo aliases found.`);
-  }
+  // const aliasData = [];
```
TateB: remove

contracts/src/resolver/interfaces/setters/IPubkeySetter.sol
```diff
@@ -0,0 +1,25 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.13;
+
+/// @dev Interface selector: `0x2b393b33`
+interface IPubkeySetter {
```
TateB: lets remove pubkey, no one's using it but if they were they could use data as a replacement

contracts/src/resolver/AbstractRecordResolver.sol
```diff
@@ -0,0 +1,224 @@
...
+
+    ////////////////////////////////////////////////////////////////////////
+    // Implementation
+    ////////////////////////////////////////////////////////////////////////
+
+    /// @notice Same as `multicall()`.
+    /// @param {node} Ignored, for interface compatibility.
+    /// @param calls The calls to make.
+    /// @return results The results of the calls.
+    function multicallWithNodeCheck(
```
TateB, line 80: what compatibility is being preserved here?

contracts/src/resolver/AbstractRecordResolver.sol
```diff
@@ -0,0 +1,224 @@
...
+    {
+        return multicall(calls);
+    }
+
+    /// @inheritdoc IMulticallable
+    /// @notice Perform multiple write operations.
+    /// @dev Reverts with first error.
+    /// @param calls The calls to make.
+    /// @return results The results of the calls.
+    function multicall(bytes[] calldata calls) public returns (bytes[] memory results) {
```
TateB, line 95: i remember we talked about keeping this for one reason or another but now i cant remember why. this doesn't seem useful?

contracts/src/resolver/interfaces/setters/IContentHashSetter.sol
```diff
@@ -0,0 +1,23 @@
...
+    event ContentHashUpdated(uint256 indexed recordId, bytes contentHash);
+
+    ////////////////////////////////////////////////////////////////////////
+    // Functions
+    ////////////////////////////////////////////////////////////////////////
+
+    /// @notice Set content hash.
+    /// @param name The DNS-encoded name.
+    /// @param contentHash The content hash.
+    function setContentHash(bytes calldata name, bytes calldata contentHash) external;
```
TateB: this is a pure nit but the getter interface considers `contenthash` all one word without camel casing so...

contracts/src/resolver/SharedResolver.sol
```diff
@@ -0,0 +1,220 @@
...
+import {IAddressSetter} from "./interfaces/setters/IAddressSetter.sol";
+import {IContentHashSetter} from "./interfaces/setters/IContentHashSetter.sol";
+import {IDataSetter} from "./interfaces/setters/IDataSetter.sol";
+import {IInterfaceSetter} from "./interfaces/setters/IInterfaceSetter.sol";
+import {INameSetter} from "./interfaces/setters/INameSetter.sol";
+import {IPubkeySetter} from "./interfaces/setters/IPubkeySetter.sol";
+import {ITextSetter} from "./interfaces/setters/ITextSetter.sol";
+
+/// @notice PublicResolver that respects the ENSv2 registry and uses name-based setters.
+contract SharedResolver is ISharedResolver, AbstractRecordResolver, DelegatedContractNamer {
```
TateB: what's the usecase for this?

TateB (changes requested):

contracts/script/testNames/display.ts
adraffy: removed

contracts/src/resolver/AbstractRecordResolver.sol
adraffy: `IMulticallable`

contracts/src/resolver/AbstractRecordResolver.sol
adraffy: I agree that if you granted hca access, it's not useful.

contracts/src/resolver/interfaces/setters/IContentHashSetter.sol
adraffy: changed, I had it the other way since the interface file is called `IContentHashResolver`

contracts/src/resolver/SharedResolver.sol
adraffy: I consider it optional: it was included as an additional example to help design the `IRecordResolver` and the setters.

It is the V2 equivalent of the `PublicResolver` that uses the new setters/events.

contracts/src/resolver/interfaces/setters/IPubkeySetter.sol
adraffy: removed

contracts/src/resolver/PermissionedResolver.sol
adraffy: Maybe `clear()` should be removed?

adraffy: removed

contracts/script/testNames.ts
```diff
@@ -218,13 +217,13 @@ export async function testNames(env: DevnetEnvironment) {
...
 
   // Show alias mappings for names that may have aliases
-  const aliasCandidates = [
-    "alias.eth",
-    "sub.alias.eth",
-    "linked.parent.eth",
-    "wallet.linked.parent.eth",
-  ];
-  await showAlias(env, aliasCandidates);
+  // const aliasCandidates = [
```
TateB: remove

hiddentao (approved):

contracts/script/testNames.ts
adraffy: removed

contracts/src/universalResolver/libraries/LibRegistry.sol
```diff
@@ -110,29 +119,51 @@ library LibRegistry {
...
+        (string memory label, uint256 next) = NameCoder.extractLabel(name, offset);
+        if (bytes(label).length == 0) {
+            return (rootRegistry, offset);
+        }
+        (registry, registryOffset) = findNearestRegistry(rootRegistry, name, next);
+        if (address(registry) != address(0)) {
+            IRegistry child = registry.getSubregistry(label);
+            if (address(child) != address(0)) {
+                registry = child;
+                registryOffset = offset;
```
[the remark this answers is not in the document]
adraffy: good spot

contracts/src/resolver/AbstractRecordResolver.sol
```diff
@@ -0,0 +1,218 @@
...
+
+    ////////////////////////////////////////////////////////////////////////
+    // Initialization
+    ////////////////////////////////////////////////////////////////////////
+
+    /// @inheritdoc ERC165
+    function supportsInterface(bytes4 interfaceId) public view virtual override returns (bool) {
+        return
+            interfaceId == type(IExtendedResolver).interfaceId ||
+            interfaceId == type(IRecordResolver).interfaceId ||
```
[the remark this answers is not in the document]
adraffy: removed interface

contracts/src/resolver/SharedResolver.sol
makoto: Let's remove SharedResolver out of this PR and  raise different PR for post mainnet release

adraffy: removed

TateB (approved):

PR #362, opened 2026-06-30
feat: add standalone hca registration and scoped sessions

TateB:
adds a standalone, single-owner HCA that can:

- register names through owner or session authorised batched calls;
- reuse one multi-chain session for either same-chain or cross-chain registration;
- pay protocol and execution costs in supported stablecoins, with registration funds pulled when the registration step is ready;
- keep resolver and primary-name updates in the same in-app session.

this also adds deterministic deployment, the minimal Rhinestone SDK patch, local and Sepolia live-proof tooling, and the HCA integration document.


contracts/src/hca/interfaces/IStandaloneHCAOwner.sol
```diff
@@ -0,0 +1,10 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.27;
+
+/// @title Standalone HCA Owner Interface
+/// @notice Minimal owner view exposed by a standalone HCA account.
+/// @dev Interface selector: `0x8da5cb5b`
+interface IStandaloneHCAOwner {
+    /// @notice Returns the account owner.
+    function owner() external view returns (address);
```
makoto, line 9: Is this the interface of `HCA.owner` mentioned at https://github.com/ensdomains/contracts-v2/pull/337/changes#diff-6a6ce90b46e67050e152fdd2989ed160e29af65d60039b521aaadffdc46d899cR116 ? If so, do we not need to change to `isOwner` so that it's not bound to single owner?

makoto:
Can you write down what sort of interface changes will be required for other components (such as https://github.com/rhinestonewtf/ens-modules and https://github.com/rhinestonewtf/sdk ) on the description section of this PR so that Rhinestone team can make the changes accordingly?

adraffy:
The Solidity code looks fine but I'm still lacking context.  I'm looking through the markdown files.

contracts/src/hca/interfaces/IStandaloneHCAOwner.sol
adraffy: I think having an interface for this (instead of reusing `Ownable`) and leaving it as single owner is ideal so there's a reliable unwrap address (to access the EOA.)

contracts/src/hca/StandaloneHCAFactory.sol
```diff
@@ -0,0 +1,89 @@
...
+    ////////////////////////////////////////////////////////////////////////
+
+    /// @notice Deploys and initializes an HCA for `owner`.
+    /// @dev The external caller does not affect the proxy address because this contract is always
+    ///      the caller observed by VerifiableFactory.
+    /// @param owner The immutable owner of the deployed HCA.
+    /// @param hcaImplementation The initial HCA implementation.
+    /// @param userSalt An optional namespace or account-version salt.
+    /// @return hca The deployed HCA proxy.
+    function deploy(address owner, address hcaImplementation, uint256 userSalt)
```
adraffy, line 57: Just checking: it is safe to create a trusted HCA for an arbitrary owner because only that account can operate the HCA?

contracts/src/hca/StandaloneSingleOwnerHCA.sol
```diff
@@ -0,0 +1,252 @@
...
+    }
+
+    /// @notice Invalidates every enabled session for this account.
+    /// @dev Increments the nonce checked by the fixed validator. Only callable by the owner
+    ///      directly; account execution paths cannot reach it because self-calls carry the
+    ///      account as `msg.sender`.
+    function revokeSessions() external {
+        if (msg.sender != _owner) {
+            revert CallerNotOwner();
+        }
```
adraffy: Convert this to a modifier?  (used 3x)

adraffy (approved):
Should `HCAOwnerAndSessionValidator` and `StandaloneHCAFactory` be nameable? (via `DelegatedContractNamer`)  This could happen in the next PR.

contracts/src/hca/HCAOwnerAndSessionValidator.sol
```diff
@@ -0,0 +1,1326 @@
...
+            result := and(mload(add(add(callData, 0x20), offset)), 0xffffffffffffffffffffffffffffffffffffffff)
+        }
+    }
+
+    /// @notice Requires an ABI address argument to match the policy-expected address.
+    /// @dev Reverts with `PolicyRuleFailed` on mismatch.
+    /// @param callData ABI-encoded call data.
+    /// @param offset Offset of the ABI word holding the address.
+    /// @param expected The only address the policy accepts at that position.
+    function _requireArgAddress(bytes memory callData, uint256 offset, address expected)
```
hiddentao, line 1929: This method isn't used anywhere

contracts/src/hca/HCAOwnerAndSessionValidator.sol
```diff
@@ -0,0 +1,1326 @@
...
+        if (_singleChainDigest(msg.sender, nonce, operationData, gasRefund) != hash) {
+            revert InvalidSessionData();
+        }
+        if (_recover(hash, data[signatureOffset:]) != sessionKey) {
+            revert InvalidSigner();
+        }
+    }
+
+    /// @dev Reconstructs the production IntentExecutor digest for a no-refund single-chain intent.
+    function _singleChainDigest(address account, uint256 nonce, bytes calldata operationData)
```
hiddentao: This 3-arg version of `_singleChainDigest` is only ever called from test-code so should we really be shipping it in production? Can't we remove/refactor this?

contracts/script/liveHcaRhinestoneRegistration.ts
```diff
@@ -0,0 +1,3602 @@
...
+  ? BigInt(process.env.HCA_CROSS_CHAIN_DESTINATION_GAS)
+  : 1_200_000n;
+const SAME_CHAIN_USDC_BUDGET = process.env.HCA_SAME_CHAIN_USDC_BUDGET
+  ? BigInt(process.env.HCA_SAME_CHAIN_USDC_BUDGET)
+  : 20_000_000n;
+const SESSION_GAS_LIMIT = process.env.HCA_SESSION_GAS_LIMIT
+  ? BigInt(process.env.HCA_SESSION_GAS_LIMIT)
+  : undefined;
+const MODE_ERC1271 = `0x0201${"00".repeat(30)}` as Hex;
+const MODE_EMISSARY_EXECUTION = `0x0204${"00".repeat(30)}` as Hex;
```
hiddentao: Where is this constant used?

contracts/script/liveHcaRhinestoneRegistration.ts
```diff
@@ -0,0 +1,3602 @@
...
+  ? BigInt(process.env.HCA_USER_SALT)
+  : 0n;
+
+const REGISTRATION_DURATION = 28n * 86400n;
+const SESSION_DURATION = 24n * 60n * 60n;
+const COIN_TYPE_ETH = 60n;
+const STATUS_REGISTERED = 2;
+const ROLES_ALL =
+  0x1111111111111111111111111111111111111111111111111111111111111111n;
+const RHINESTONE_INTENT_EXECUTOR =
```
hiddentao: why not import `RHINESTONE_INTENT_EXECUTOR` and `RHINESTONE_GAS_REFUND_PAYMASTER` from `deploy-constants`?

contracts/script/checkLiveHcaRegistration.ts
```diff
@@ -0,0 +1,411 @@
...
+  return env;
+}
+
+function parseSummary(stdout: string): LiveRegistrationSummary {
+  const trimmed = stdout.trim();
+  const start = trimmed.lastIndexOf("\n{");
+  return JSON.parse(start === -1 ? trimmed : trimmed.slice(start + 1));
+}
+
+const proc = Bun.spawn(["bun", "script/liveHcaRhinestoneRegistration.ts"], {
```
hiddentao: Seems very similar code to that in `checkLiveHcaRhinestonePermitPull.ts` - why not refactor and reuse this harness code?

hiddentao (changes requested):

contracts/src/hca/StandaloneHCAFactory.sol
TateB: the hca implementation is an arbitrary address. protocol contracts and frontends will only trust implementations that are known to enforce single owner.

PR #363, opened 2026-07-01
Add `ETHHelper.getState()`

adraffy:
- added `ETHHelper`
    * added `getState(label)`


Arachnid:
Can you elaborate on the need for this?

adraffy:
This doesn't need to be in the audit.  This is a helper for frontend.  

1 RPC call gives you the complete `{label}.eth` information.

contracts/deploy/01_ETHHelper.ts
```diff
@@ -0,0 +1,29 @@
...
+        nameWrapper.address,
+        rootRegistry.address,
+        contractNamer.address,
+        GRACE_PERIOD_V2,
+      ],
+    });
+  },
+  {
+    tags: ["ETHHelper", "v2"],
+    dependencies: ["NameWrapper", "RootRegistry", "ContractNamer"],
```
makoto: Can you add `ETHRegistry` as a dependency so that this runs after ETHRegistry is deployed?  Otherwise `ETH_REGISTRY = rootRegistry.getSubregistry("eth") ` may return 0x for new deployment during mainnet release

contracts/deploy/01_ETHHelper.ts
```diff
@@ -0,0 +1,29 @@
...
+  },
```
makoto, line 33: Also needs `migration:phase1:deploy-v2 ` tag . Confirm with @hiddentao

contracts/src/universalResolver/ETHHelper.sol
```diff
@@ -0,0 +1,147 @@
...
+        if (state.status == IPermissionedRegistry.Status.RESERVED) {
+            ethState.expiry = uint64(BASE_REGISTRAR.nameExpires(labelId));
+            bytes32 node = NameCoder.namehash(NameCoder.ETH_NODE, bytes32(labelId));
+            if (NAME_WRAPPER.isWrapped(NameCoder.ETH_NODE, bytes32(labelId))) {
+                ethState.tokenId = uint256(node);
+                ethState.owner = NAME_WRAPPER.ownerOf(uint256(node));
+                ethState.status = Status.WRAPPED;
+            } else {
+                ethState.tokenId = labelId;
+                try BASE_REGISTRAR.ownerOf(labelId) returns (address owner) {
```
makoto: Probably this is better because `try BASE_REGISTRAR.ownerOf(labelId) ` could swallow any other unexpected errors
```
    if (ethState.expiry > block.timestamp) {
        ethState.owner = BASE_REGISTRAR.ownerOf(labelId);   // cannot revert on expiry now
        ethState.status = Status.UNWRAPPED;
    } else {
        ethState.status = Status.GRACE_V1;
        ethState.expiry = state.expiry;
    }
```

makoto (changes requested):

contracts/deploy/01_ETHHelper.ts
adraffy: changed

contracts/deploy/01_ETHHelper.ts
adraffy: changed

contracts/src/universalResolver/ETHHelper.sol
adraffy: I think they're equivalent but changed

makoto (approved):

PR #382, opened 2026-07-27
fix(migration): revoke superseded deployments' v1 authorizations

hiddentao:
Phase 3 previously only checked the four named v1 registration controllers on the BaseRegistrar. A re-deploy left the prior deployment's handoff contracts authorized, including the permissionless TestnetV1PremigrationRegistrar, silently reopening v1 .eth registration into an archived v2 registry.

The freeze now audits the BaseRegistrar and both reverse registrars, drawing candidates from the registration controllers, the handoff contracts of every deployment namespace on the chain, and the registrar's ControllerAdded history, and revokes everything the active deployment did not itself grant. verify-v1-registrars-disabled asserts that same complete set and fails on anything outside it. When the controller history cannot be read both commands warn and fall back to artifact-backed candidates.

fork full reclaims registrar ownership before the freeze rather than after it, and re-runs the verification once the phase 6 handoff grants land.

### DIFF
--- a/contracts/script/migration.ts
+++ b/contracts/script/migration.ts
@@ -23,6 +23,7 @@ import {
   http,
   keccak256,
   namehash,
+  parseAbiItem,
   parseEther,
   stringToHex,
   zeroAddress,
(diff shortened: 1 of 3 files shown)


contracts/script/migration.ts
```diff
@@ -1561,94 +1561,468 @@ async function sendAdminWrite(opts: {
...
+// result is never mistaken for a clean controller set.
+async function discoverV1ControllerAddresses(
+  client: ReturnType<typeof publicClient>,
+  address: Address,
+): Promise<Address[] | null> {
+  try {
+    const toBlock = await client.getBlockNumber();
+    const logs = await Promise.all(
+      [V1_CONTROLLER_ADDED_EVENT, V1_CONTROLLER_REMOVED_EVENT].map((event) =>
+        client.getLogs({ address, event, fromBlock: 0n, toBlock }),
```
makoto: Is getting log from block 0 feasible in public net (both sepolia and mainnet)?

contracts/script/migration.ts
```diff
@@ -1561,94 +1561,468 @@ async function sendAdminWrite(opts: {
...
-      return controller ? [{ name, address: controller.address }] : [];
-    }),
-    ...(opts.extraControllers ?? []),
+  const registrarCandidates = candidatesFor(V1_BASE_REGISTRAR_NAME);
+  for (const name of V1_REGISTRATION_CONTROLLER_NAMES) {
+    const controller = loadV1Deployment(opts.network, name, opts);
+    if (controller) addCandidate(registrarCandidates, controller.address, name);
+  }
+
+  for (const namespace of siblingDeploymentNamespaces({
```
makoto: Is there a risk that siblingDeploymentNamespaces returns nothing (due to chain id mismatch or non prefixed name) then skipping line 1864 loop?

contracts/script/migration.ts
```diff
@@ -1561,94 +1561,468 @@ async function sendAdminWrite(opts: {
...
+] as const;
+
+const V1_BASE_REGISTRAR_NAME = "BaseRegistrarImplementation";
+
+// v1 reverse registrars a migration is granted control of: the premigration
+// registrar writes reverse records on a registrant's behalf, and the adapters
+// forward reverse updates for the accounts they are allowed to name. These are
+// shared v1 contracts, so a superseded deployment's grant stays live across
+// re-deploys.
+const V1_REVERSE_REGISTRAR_NAMES = [
```
makoto, line 1587: Would there be a risk of depending on locally deployed artifacts to fetch the contract detail over getting from `ControllerChanged` events?

contracts/script/migration.ts
hiddentao: It's ok with a good RPC

contracts/script/migration.ts
hiddentao: Will strengthen this logic, thanks!

contracts/script/migration.ts
hiddentao: A small one, will strengthen this. Thanks.

makoto (approved):
A minor point is that the quick start command will fail without `--save-deployments` and `--ur-manager` flag so may want to update the doc

PR #386, opened 2026-07-31
Fix critical HCA spoof vulnerabilities

TateB:
this fixes the two critical HCA spoofing paths identified during review: runtime implementation/owner spoofing against the reverse adapters, and inherited Nexus delegatecall execution allowing temporary HCA storage spoofing.

changes:

- made `StandaloneHCAFactory` ownable, with owner-controlled approval of initial implementations and immutable deployment-time owner bindings. reverse adapters now authorise against that factory record instead of mutable `verifyContract()` / `owner()` state during the call
- kept `VerifiableFactory` for proxy deployment and upgrade verification, but removed it from runtime HCA identity authorisation
- prohibited HCA delegatecall execution, which allowed delegated code to temporarily overwrite the HCA owner slot and impersonate another account to the reverse adapters. the restriction covers both installed-executor calls and owner-signed user-ops while retaining single and batch call execution
- removed the separate `TrustedHCASet` and updated deployment/live tooling for the new implementation approval flow
- updated reverse-registrar adapter deployments to revoke controller access from replaced adapters, including across interrupted deployment resumes
- added working exploit PoCs and regression coverage for both spoof paths, plus deployment-resume e2e coverage

notes:

- the mutable Rhinestone intent executor remains a separate mainnet blocker: every HCA permanently trusts it with arbitrary call/batch authority, so a malicious executor upgrade could spend HCA-held assets or exercise its authorisations without an owner or session signature


contracts/test/unit/reverse-registrar/DefaultReverseRegistrarAdapter.t.sol
```diff
@@ -182,4 +183,125 @@ contract DefaultReverseRegistrarAdapterTest is HCAFixture {
...
+    }
+}
+
+
+/// @title Transiently Spoofable Default Adapter
+/// @notice Reproduces implementation-and-owner checks without factory deployment provenance.
+contract TransientlySpoofableDefaultAdapter {
+    DefaultReverseRegistrar internal immutable REGISTRAR;
+    IVerifiableFactory internal immutable VERIFIABLE_FACTORY;
+    IAddressSet internal immutable TRUSTED_HCA_SET;
```
adraffy, line 273: Should this still be using `TRUSTED_HCA_SET` design?

contracts/src/hca/StandaloneSingleOwnerHCA.sol
```diff
@@ -218,6 +275,19 @@ contract StandaloneSingleOwnerHCA is Nexus, IProxyAuthorization {
             predecessorGate.approvedImplementations(previousImplementation);
```
adraffy, line 275: Gate -> `IAddressSet`?

contracts/test/fixtures/HCAFixture.sol
```diff
@@ -5,61 +5,91 @@ import {Test} from "forge-std/Test.sol";
...
+        address hca = verifiableFactory.deployProxy(hcaImpl, nextSalt++, data);
+        standaloneHCAFactory.recordDeployment(hca, hcaOwner);
+        return hca;
+    }
+}
+
+
+/// @title Mock HCA Deployment Registry
+/// @notice Records immutable owner bindings treated as factory certifications by adapter tests.
+contract MockHCADeploymentRegistry {
```
adraffy: Why isn't this just a `StandaloneHCAFactory`?

adraffy (changes requested):
Moving the verification checks to deployment and internalizing the owner mapping makes sense.

contracts/test/unit/reverse-registrar/DefaultReverseRegistrarAdapter.t.sol
TateB: this is a regression test, not current design

contracts/src/hca/StandaloneSingleOwnerHCA.sol
TateB: resolved in #390

adraffy (approved):

PR #391, opened 2026-08-04
fix: pin Solidity compiler profiles

TateB:
pins deployable contracts to their intended compiler versions and adds regression coverage for compiler settings. without this, adding the HCA compiler allowed hardhat to silently rebuild compatible protocol contracts with it.

### DIFF
--- a/contracts/test/mocks/MockStandaloneHCAStack.sol
+++ b/contracts/test/mocks/MockStandaloneHCAStack.sol
@@ -1,12 +1,18 @@
 // SPDX-License-Identifier: MIT
 pragma solidity ^0.8.27;
 
+import {IProxyAuthorization} from "@ensdomains/verifiable-factory/IProxyAuthorization.sol";
+import {UUPSUpgradeable} from "@openzeppelin/contracts-upgradeable/proxy/utils/UUPSUpgradeable.sol";
 import {PackedUserOperation} from "account-abstraction/interfaces/PackedUserOperation.sol";
 import {IExecutor} from "nexus/interfaces/modules/IExecutor.sol";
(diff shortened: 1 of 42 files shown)


contracts/test/mocks/MockStandaloneHCAStack.sol
```diff
@@ -59,6 +65,78 @@ contract MockStandaloneHCA {
 }
 
 
+/// @title Mock Address Set
+/// @notice Test-only administrator-controlled address set for HCA upgrade authorization.
+contract MockAddressSet is IAddressSet {
```
adraffy: This is just a `PermissionedAddressSet` ?

contracts/test/mocks/MockStandaloneHCAStack.sol
```diff
@@ -1,12 +1,18 @@
 // SPDX-License-Identifier: MIT
 pragma solidity ^0.8.27;
```
adraffy: Pin

contracts/test/mocks/MockStandaloneHCAStack.sol
```diff
@@ -59,6 +65,78 @@ contract MockStandaloneHCA {
...
+    /// @inheritdoc IAddressSet
+    function includes(address member) external view returns (bool) {
+        return _approved[member];
+    }
+}
+
+
+/// @title Mock HCA Registration Resolver
+/// @notice Test-only upgradeable resolver target implementing the HCA policy call surface.
+contract MockHCARegistrationResolver is
```
adraffy: This is just a `PermissionedResolver`?

adraffy (changes requested):

adraffy (approved):

PR #392, opened 2026-08-04
fix: authorize HCA registrars through registry roles

TateB:
pins the HCA validator to `ETHRegistry` instead of one `ETHRegistrar`, and authorises registrar calls and payment approvals through the live root `ROLE_REGISTRAR`. this allows registrar replacement or multiple active registrars without upgrading every HCA, while role revocation takes effect immediately.

### DIFF
--- a/contracts/test/mocks/MockStandaloneHCAStack.sol
+++ b/contracts/test/mocks/MockStandaloneHCAStack.sol
@@ -8,6 +8,28 @@ import {EncodedModuleTypes} from "nexus/lib/ModuleTypeLib.sol";
 
 import {HCAOwnerAndSessionValidator} from "~src/hca/HCAOwnerAndSessionValidator.sol";
 
+/// @title Mock Registrar Role Registry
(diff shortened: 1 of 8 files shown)


contracts/test/mocks/MockStandaloneHCAStack.sol
```diff
@@ -8,6 +8,28 @@ import {EncodedModuleTypes} from "nexus/lib/ModuleTypeLib.sol";
 
 import {HCAOwnerAndSessionValidator} from "~src/hca/HCAOwnerAndSessionValidator.sol";
 
+/// @title Mock Registrar Role Registry
+/// @notice Test-only registry stand-in that exposes mutable root-role membership.
+contract MockRegistrarRoleRegistry {
```
adraffy, line 13: I think you should just use a `PermReg`

adraffy (changes requested):
The `ethRegistry.hasRootRoles()` check makes sense.

adraffy (approved):

PR #409, opened 2026-08-10
Add mock usdc as secondary only for sepolia

makoto:
## Summary

Sepolia HCA registrations can now pay with the faucet-minted MockUSDC: the
validator's secondary payment-token slot, previously duplicating production
USDC, defaults to `SEPOLIA_MOCK_USDC`. Includes the redeployed validator and
`StandaloneHCAImplementation` artifacts.
This is requested by frontend team so that they can auto mint test usdc to users.

NOTE: We hardcoded a constant rather than using the existing
`HCA_SECONDARY_PAYMENT_TOKEN` env override: this is Sepolia's permanent
default, not an operator override — env-only config isn't recorded in the
repo, so any future redeploy without it would silently revert the slot to
USDC. The constant also pins the correct MockUSDC (the live oracle rejects
the archived June one) and is covered by tests.


contracts/test/unit/reverse-registrar/ReverseRegistrarAdapter.t.sol
```diff
@@ -159,11 +159,7 @@ contract ReverseRegistrarAdapterTest is HCAFixture {
 
         vm.expectRevert(abi.encodeWithSelector(HCAAuthorizer.HCADeploymentNotTrusted.selector, hca));
         vm.prank(hca);
-        bytes32 node = reverseAdapter.claimWithHCA(owner, resolver);
```
TateB, line 162: ?

makoto: It looks like your original  `test_claimWithHCA_revert_hcaDeploymentNotTrusted`  test at https://github.com/ensdomains/contracts-v2/pull/386/changes#diff-920b80acd20d3d2a52dbe4101847267caa69b8341e4563b7c76d409bd59446cdR152 didn't have these assertions, but it was brought in when resolving some merge conflict. 
You can see that the test on  `feat/verifiable-reverse-adapters` doesn't have these assertion either. https://github.com/ensdomains/contracts-v2/pull/337/changes#diff-920b80acd20d3d2a52dbe4101847267caa69b8341e4563b7c76d409bd59446cdR152 . You can revert the commit if you now think it is necessary but otherwise I suggest leaving as is

TateB (approved):

PR #420, opened 2026-08-24
refactor: extract shared HCA validator libraries

TateB:
the two HCA validators had overlapping decoding/validation code that could be shared between them.

changes:
- moved repeated validation logic into shared HCA libraries
- kept the validators' existing behavior and external interfaces


contracts/src/hca/libraries/HCAExecutionLib.sol
```diff
@@ -0,0 +1,65 @@
...
+
+    /// @notice Copies function arguments out of encoded call data.
+    /// @dev Reverts when the call data does not contain a complete selector.
+    /// @param callData ABI-encoded call data with a function selector prefix.
+    /// @return args ABI-encoded function arguments.
+    function callArgs(bytes memory callData) internal pure returns (bytes memory args) {
+        if (callData.length < 4) {
+            revert InvalidOperationEncoding();
+        }
+        args = new bytes(callData.length - 4);
```
adraffy:
```
uint256 size =callData.length - 4;
args = new bytes(size);
assembly {
    mcopy(add(args, 36), add(callData, 36), size)
}
```

contracts/src/hca/libraries/HCARegistrarPolicyLib.sol
```diff
@@ -0,0 +1,98 @@
...
+    {
+        if (operationData.length < 32) {
+            return false;
+        }
+        Execution[] memory executions = abi.decode(operationData[32:], (Execution[]));
+        for (uint256 i; i < executions.length; ++i) {
+            if (HCAExecutionLib.selector(executions[i].callData) != IETHRegistrar.register.selector) {
+                continue;
+            }
+            if (!isPaymentToken(executions[i].target, token)) {
```
adraffy, line 78: since this function is batched, would it make sense to extract the `IRentPriceOracle` once?

contracts/src/hca/libraries/HCAResolverPolicyLib.sol
```diff
@@ -0,0 +1,184 @@
...
+        grants[0] = Grant({account: account, roleBitmap: EACBaseRolesLib.ALL_ROLES});
+        grants[1] = Grant({account: owner, roleBitmap: EACBaseRolesLib.ALL_ROLES});
+        bytes[] memory calls = new bytes[](0);
+        bytes memory expectedInitData =
+            abi.encodeCall(IPermissionedResolverInitializable.initialize, (grants, calls));
+        bytes memory expectedCallData =
+            abi.encodeCall(IVerifiableFactory.deployProxy, (implementation, salt, expectedInitData));
+
+        if (
+            keccak256(callData) != keccak256(expectedCallData) ||
```
adraffy, line 136: I considered relaxing this to just hashing the first argument (`Grants[]`) but I think it's fine as-is since the frontend deployment can populate the resolver in later steps.

contracts/src/hca/libraries/HCAResolverPolicyLib.sol
```diff
@@ -0,0 +1,184 @@
...
+    function _checkCalls(bytes[] memory calls, address owner) private pure {
+        for (uint256 i; i < calls.length; ++i) {
+            checkCall(calls[i], owner);
+        }
+    }
+
+    /// @dev Returns whether a selector is a resolver record setter accepted by the policy.
+    function _isRecordSelector(bytes4 callSelector) private pure returns (bool) {
+        return
+            callSelector == IPermissionedResolver.linkToRecord.selector ||
```
adraffy, line 174: I also was unsure if this should include grant/revoke.

contracts/src/hca/libraries/HCASignatureLib.sol
```diff
@@ -0,0 +1,127 @@
...
+        }
+
+        bytes32 r;
+        bytes32 s;
+        uint8 v;
+        assembly ("memory-safe") {
+            r := calldataload(signature.offset)
+            s := calldataload(add(signature.offset, 0x20))
+            v := byte(0, calldataload(add(signature.offset, 0x40)))
+        }
```
adraffy: Refactor to `rsvFromSignature(bytes): (r s, v)` and reuse in `recover()`?

adraffy (changes requested):
The refactor into libs was a good idea.

contracts/src/hca/libraries/HCAExecutionLib.sol
TateB: noting args+32 not 36

contracts/src/hca/libraries/HCARegistrarPolicyLib.sol
TateB: current behaviour allows for multiple registrars (with potentially different oracles) but agreed this can be simplified (just enforce one registrar), adding to #426

contracts/src/hca/libraries/HCASignatureLib.sol
TateB: added

adraffy (approved):

PR #422, opened 2026-08-24
Fix DNS deployment

adraffy:
- removed `01_DNSV1Mirror`
- included `DNSTLDResolver`, `DNSTXTResolver`, and `DNSAliasResolver` in deployment
- ensure suffixes are only TLDs


contracts/deploy/01_DNSTLDResolver.ts
```diff
@@ -55,18 +55,23 @@ export default execute(
       ],
     });
 
-    const candidates = tags.local
+    if (tags["clean-testnet"]) {
+      console.warn("  - Skipping v1 DNS TLD registration on clean-testnet");
+      return;
```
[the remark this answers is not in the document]
adraffy: fixed by removing this option

contracts/deploy/01_DNSTLDResolver.ts
```diff
@@ -55,18 +55,18 @@ export default execute(
       ],
     });
 
-    const candidates = tags.local
+    const tlds = tags.local
       ? ["com", "org", "net", "xyz"]
-      : await fetchPublicSuffixes();
+      : await fetchPublicSuffixes(true);
```
hiddentao: Is this method ever called without `true`? if not then we can remove the argument and simplify the method itself.

contracts/script/publicSuffixes.ts
```diff
@@ -1,6 +1,6 @@
 import { labelhash, zeroAddress, type Abi, type Address } from "viem";
 import { dnsEncodeName } from "../test/utils/utils.js";
-import { MAX_EXPIRY, ROLES } from "./deploy-constants.js";
+import { MAX_EXPIRY, ROLES, STATUS } from "./deploy-constants.js";
 
 export const SUFFIX_BATCH_SIZE = 25;
```
hiddentao, line 5: Where is this used outside of this file?

hiddentao (changes requested):

contracts/deploy/01_DNSTLDResolver.ts
adraffy: I wasn't sure on the design.   `PublicSuffixList.isPublicSuffix(bytes name)` can be an arbitrary name and the public suffix data file contains domains of various depths, not just TLDs.

The filter could be moved to `const tlds = (...).filter(x => !x.includes("."));` ?

contracts/script/publicSuffixes.ts
adraffy: It's used a bunch of places, mostly `scripts/` and `tests/`.  It's the JS equivalent of the Solidity enum.  I only changed it to make the `status === 0` more clear.

contracts/deploy/01_DNSTLDResolver.ts
adraffy: Changed to inline.

hiddentao (approved):

PR #434, opened 2026-08-31
Add contract naming scripts

adraffy:
wip

---

- added `contract-names.json`
- refactored devnet to use those names
- moved salt logic to `script/salts.ts`
- ~~added deploy script~~


makoto:
Should adding versioning logic (eg: v1.root.registry.ens.eth with. inode alias) be included here or it's more of the role of the deployment script?

PR #440, opened 2026-09-08
Fix `findResolver()` not checking `IExtendedResolver`

adraffy:
**Security issue:**  by convention (see: [ens-contracts](https://github.com/ensdomains/ens-contracts/blob/staging/contracts/universalResolver/RegistryUtils.sol)), `findResolver()` returns the "raw" resolver result without validity checks.  In ENSv1, `UniversalResolver` was the primary consumer and it enforced explicit validity checks via `_checkResolver()`.  In ENSv2, there are multiple contracts that invoke resolvers by name (`AbstractMirrorResolver`, `ENSV1Resolver`, `ENSV2Resolver`, `DNSTLDResolver`, `DNSAliasResolver`).  They all use `ResolverCaller` which reverts `UnreachableName` if the resolver is codeless but are unable to enforce [ENSIP-10](https://docs.ens.domains/ensip/10/#specification) `IExtendedResolver` check (which requires knowledge of how the resolver was found.)

---

- changed `LibRegistry` to `LibResolution`
    * renamed `findResolver()` to `findUnvalidatedResolver()`
    * added `validateResolver(address, bool foundAtLeaf): address` to check `IExtendedResolver`
    * readded `findResolver()` which calls `findUnvalidatedResolver() + validateResolver()`

### DIFF
--- a/contracts/script/migration.ts
+++ b/contracts/script/migration.ts
@@ -1375,6 +1375,17 @@ async function verifyPreMigration(opts: {
   let verifiedActive = 0;
   let verifiedExpiredBonus = 0;
 
+  // reduce typescript burden
+  const getStateAbi = Artifact_PermissionedRegistry.abi.filter(
+    (x) => x.type === "function" && x.name === "getState",
+  );
+  const nameExpiresAbi = Artifact_BaseRegistrarImplementation.abi.filter(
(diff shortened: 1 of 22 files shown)


makoto:
@adraffy can you put the summary of the security vulnerability which lead to this fix?

TateB (approved):

